PR #2, opened 2023-09-12
Add convenience nvmrc for using correct Node version

ALRubinger:


packages/protocol/.nvmrc
```diff
@@ -0,0 +1 @@
+v20.3.0
```
KendallWeihe, line 1: This is inside the `protocol` package, but I imagine we would want to use the same node version across all three packages, yeah?

ALRubinger: Updated PR

KendallWeihe (approved):

ALRubinger:
Ready to go - only to include the root-level `nvmrc`; docs updates are handled unified in #16

PR #18, opened 2023-09-12
Always pass string[] for get exchanges

phoebe-lew:


packages/http-server/src/types.ts
```diff
@@ -25,7 +25,7 @@ export type GetOfferingsFilter = {
 }
 
 export type GetExchangesFilter = {
-  exchangeId?: string | string[]
+  exchangeIds?: string[]
```
mistermoe, line 28: @phoebe-lew i think we have to keep this as `exchangeId`, because: `?exchangeId=1&exchangeId=2&exchangeId=3`

phoebe-lew: ohhh thx

PR #30, opened 2023-09-25
`pnpm` and `changesets`

mistermoe:
# Changes

## Switched from `npm` -> `pnpm`

> [!NOTE] 
> learn about `pnpm` [here](https://pnpm.io).

### Motivations behind the switch

[better support](https://pnpm.io/workspaces) for multi-package repos.

---

`pnpm` prevents phantom dependencies. A "phantom dependency" occurs when a project uses a package that is not defined in its package.json file. 

how do these occur? When installing dependencies with npm, all packages are hoisted to the root of the modules directory. As a result, source code has access to dependencies that are not added as dependencies to the project which means you can `import` something that you haven't included as a dependency in `package.json`

By default, pnpm uses symlinks to add only the direct dependencies of the project into the root of the modules directory

---

Human readable lock files. `package-lock.json` merge conflicts feel like an impossible battle to win. `pnpm-lock.yaml` is far easier to read (at least imo)

---

`pnpm` includes a command (`pnpm licenses list`) that can be used to list all licenses used by all dependencies. this is especially helpful when building open source libraries

---

the above aren't even the primary motivations behind why [other projects](https://pnpm.io/workspaces#usage-examples) are switching. if you're interested in those you can read about them [here](https://pnpm.io/motivation)

## Introduced `changesets`

> [!NOTE]
> learn about `changesets` [here](https://github.com/changesets/changesets/tree/main#readme). 

managing the versions of all 3 packages in this repo by hand was getting supa cumbersome and eagerly waiting for the opportunity to ruin someone's day. 

Decided to give `changesets` a shot to handle version management for us! i've configured `changesets` to do [fixed versioning](https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md). This means that packages will be version-bumped and published together even when there are no changes done to some of the member packages. 

Given that:
* `http-server` depends on `http-client` and `protocol` 
* `http-client` depends on `protocol`

any changes to `protocol` or `http-client` would end up version bumping all 3 packages anyway. 

I've also installed the [`changesets` bot](https://github.com/changesets/bot) into this repo. It'll comment on PRs saying that either: 
* a changeset might need to added (note that PRs changing things like documentation generally don't need a changeset) 
* say that the PR is good and already has a changeset.

# Remaining TODOs
* setting up [changesets github action](https://github.com/changesets/action)
* adding docs to `README` explaining how to use `changesets` manually


phoebe-lew (approved):

michaelneale:
Have used it for a while with dev site and it was fine there so +1 from me

.changeset/hot-worms-kneel.md
jiyoontbd, line 1: are these files under .changeset/ supposed to have odd names?

jiyoontbd (approved):

.github/workflows/integrity-check.yml
```diff
@@ -11,69 +11,82 @@ on:
...
-      - name: Install latest npm
-        run: npm install -g npm@latest
-
       - name: Install dependencies
-        run: npm ci
+        run: pnpm install
 
       - name: Build all workspace packages
-        run: npm run build
+        run: pnpm run build
```
KendallWeihe, line 59: If I'm understanding the docs right, there is no `pnpm run` (tho perhaps an alias exists) so instead this is just... 

```suggestion
        run: pnpm build
```

KendallWeihe:
Nice, yeah this looks solid AF. The description is well thought-out too, so thank you for that 👍 

I'm slightly sus that an automated changebot is something we'll want, but won't know for sure until we get further into the experience

.github/workflows/integrity-check.yml
KendallWeihe: okay yeah getting my hands dirty it must support an alias, but I would recommend we go with [the official docs](https://pnpm.io/pnpm-cli#commands) and remove the unnecessary `run` command

KendallWeihe: Done in [commit below](https://github.com/TBD54566975/tbdex-js/pull/30/commits/02e0e357aa39c285758d65485f95b6d83ec68c41)

.changeset/hot-worms-kneel.md
KendallWeihe: Okay I think I understand how this works...

- The bot can detect if an update is needed
  - It can comment on the PR to remind the developer know to do their duty
  - Also, somehow it can detect patch/minor/major??? Which I'd be very curious to understand how it is able to detect that
- Then you run `pnpm changeset`
  - This automatically generates these [funnily-named](https://github.com/changesets/changesets/blob/main/docs/common-questions.md#each-changeset-is-its-own-file) markdown files
  - I think there's a parallel here to staging a commit using `git add` prior to the actual commit
- Then you run `pnpm changeset version` 
  -  This actually executes the version update
  - Which, appends the changeset to the relevant CHANGELOG.md files
  - Updates the version in the relevant package.json files
  - And lastly, deletes the funnily-named changeset files

So the point is, these funnily-named files are ephemeral, and only exist in between the `changeset` (again, akin to "staging" a commit using `git add`) and `changeset version` (which executes the changeset update)

mistermoe:
Re: changebot

Agreed! Very interested to see how it plays out. 

Oh yeah, as an alternative to figuring out CI in this PR, we can just run `pnpm changeset version` to have `changeset` version bump everything based on the two changesets present. Then, commit / push the outcome of running that. 

Let's us tackle CI stuff in a separate PR. 


Down with either route!

.changeset/hot-worms-kneel.md
KendallWeihe: Okay, and so I think the idea of the [GitHub Action](https://github.com/changesets/action) is...

1. As developers open PRs, the changeset bot will comment reminding them to update the version
2. So then they'll run `pnpm changeset`
3. This will create a new funnily-named markdown file 
4. Then they will merge their PR (without running `changeset version` -- meaning, all those funnily-named markdown files will be merged into main)
5. At which point, the GitHub Action will automatically open a new Pull Request 
6. This new PR will automatically run the `changeset version` such that the package.json files & the CHANGELOG.md files are updated
7. Then, let's say a second developer comes in with a new PR and merges it into main
8. That PR which the GitHub Action previously opened (from step 5) will automatically re-run the `changeset version` to pull in the new changes
9. At some point in the future, the team is ready for a new release, at which point, they simply merge the open PR

KendallWeihe: So to sum up, developers stage their changesets on each PR, and the GitHub Action automates the execution of the changeset & the resulting release

KendallWeihe:
Manually executed `changeset version`, created new issues for the two remaining TODOs

- https://github.com/TBD54566975/tbdex-js/issues/31
- https://github.com/TBD54566975/tbdex-js/issues/32

Going ahead and merging this and will do the other two things in follow-up PR(s)

.github/workflows/integrity-check.yml
```diff
@@ -11,47 +11,58 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+env:
+  CI: true
```
KendallWeihe, line 15: What is `CI: true` doing here?

KendallWeihe: Interesting, I see it here https://pnpm.io/using-changesets

I also see explicit execution of `pnpm config set store-dir $PNPM_CACHE_FOLDER` which we aren't doing in our jobs

Issue #31, opened 2023-09-27
Add changesets GitHub Action


KendallWeihe:
To automate changeset version execution add the changeset GitHub Action https://github.com/changesets/action

KendallWeihe:
@mistermoe thoughts on automating package publication? Or should that always be a manual step?

KendallWeihe:
Tested integration here https://github.com/TBD54566975/tbdex-js/actions/runs/6341704162

You can see the automated PR opened here https://github.com/TBD54566975/tbdex-js/pull/36

All appears well ✅

KendallWeihe:
Completed

leordev:
@KendallWeihe beautiful! I want to add the tbdocs reporter + generator on the release PR.

Is there a reason why the PR #36 didn't trigger the CI action? 🤔

KendallWeihe:
> Is there a reason why the PR #36 didn't trigger the CI action?

@leordev Are you referring to the changeset action? It's because there weren't any new markdown files in the `.changeset/` directory.

As for the release (a different CI action), right now it's still a manual step. We may consider automating that, but that would affect the governance of change management so we'll need to think it through. For now, at least we can automate the semantic versioning.

PR #33, opened 2023-09-28
Add changesets GitHub Action #31

KendallWeihe:


KendallWeihe:
Inspiration from https://github.com/changesets/action#without-publishing

PR #35, opened 2023-09-28
Patch get-func-name audit failure

KendallWeihe:


KendallWeihe:
Tested out the bot auto-detect and it didn't pick up what should happen here, which is a patch

PR #38, opened 2023-10-03
Setup TBDocs

leordev:
Install [tbdocs-action](https://github.com/TBD54566975/tbdocs) which will have two main functionalities:

1. **Docs Reporter**: every PR against main will report the doc warnings and errors in the PR comment. The developer can feel free to correct them if they want to or ignore it for now. After releasing the v1 we should make this blocking.

2. **Publishing Docs to the developer.tbd.website**: on the Publish workflow now you have three options:
a. Publish npm: like it is today
b. Publish docs: will run the docs reporter (from step 1) and if it completes without errors it will generate the markdown files and open a PR to our website repo: https://github.com/TBD54566975/developer.tbd.website
c. Publish both: this will try to publish the docs by opening the PR, if it succeeds it will publish to npm, otherwise it will fail.

Proposed design can be found here: [TBD OSP Automated Docs Pipeline Proposal](https://hackmd.io/@leordev/Byf3R5_J6)

I've already addressed the points raised from @KendallWeihe on my forked repo example [here](https://github.com/leordev/tbdex-js/pull/2)

And this is what we get automagically:
- Automated PR: https://github.com/TBD54566975/developer.tbd.website/pull/796
- Preview of the docs: https://deploy-preview-796--tbd-website-developer.netlify.app/docs/tbdex/tbdex-js/protocol/

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -32,43 +32,46 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
 
-      - name: install pnpm
-        uses: pnpm/action-setup@v2
-        with:
-          version: 8
+      - name: test
+        run: echo "success"
 
-      - name: Set up Node.js
-        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
-        with:
-          node-version: 20
-          registry-url: https://registry.npmjs.org/
+      # - name: install pnpm
+      #   uses: pnpm/action-setup@v2
+      #   with:
+      #     version: 8
 
-      - name: Install dependencies
-        run: pnpm install
+      # - name: Set up Node.js
+      #   uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+      #   with:
+      #     node-version: 20
+      #     registry-url: https://registry.npmjs.org/
 
-      - name: Build all workspace packages
-        run: pnpm build
+      # - name: Install dependencies
+      #   run: pnpm install
 
-      - name: TBDocs Reporter and Generator
-        id: tbdocs-reporter-generator
-        uses: TBD54566975/tbdocs@main
-        with:
-          bot_app_id: ${{ secrets.TBDOCS_BOT_APP_ID }}
-          bot_app_private_key: ${{ secrets.TBDOCS_BOT_APP_PRIVATE_KEY }}
-          bot_app_installation_id: ${{ secrets.TBDOCS_BOT_APP_INSTALLATION_ID }}
-          project_path: "./packages/protocol"
-          docs_reporter: api-extractor
-          fail_on_error: true
-          fail_on_warnings: false
-          docs_generator: 'typedoc-markdown'
-          docs_target_owner_repo: 'TBD54566975/developer.tbd.website'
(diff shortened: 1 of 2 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -16,7 +25,50 @@ permissions:
   id-token: write # necessary for NPM provenance
 
 jobs:
+  tbdocs-publish:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.inputs.jobSelector == 'all' || github.event.inputs.jobSelector == 'docs' }}
```
KendallWeihe, line 30: What's the intent of this line? I get it's to control when the job runs, but where would we set the value to `all` or `docs`?

.github/workflows/npm-publish.yml
```diff
@@ -16,7 +25,50 @@ permissions:
...
+          fail_on_warnings: false
+          docs_generator: 'typedoc-markdown'
+          docs_target_owner_repo: 'TBD54566975/developer.tbd.website'
+          docs_target_branch: 'tbdocs-bot/tbdex-js/protocol'
+          docs_target_pr_base_branch: 'main'
+          docs_target_repo_path: 'site/docs/tbdex/tbdex-js/protocol'
+
   publish-npm:
+    needs: tbdocs-publish
+    if: ${{ (github.event.inputs.jobSelector == 'all' && needs.tbdocs-publish.result == 'success') || (github.event.inputs.jobSelector == 'npm' && always())}}
```
KendallWeihe, line 71: in the same vein as the previous question, when is the value set to `npm`? I'm not familiar with the way GitHub Actions works here

is `always()` strictly needed here?

KendallWeihe (approved):

.github/workflows/npm-publish.yml
leordev: I just used `docs` to create the preview of our functionality in this PR.

I'd imagine we should always use `all` when cutting new releases, to force us to keep up with good doc quality.

But if we are in a rush, we can simply just run `npm` now and keep working on the docs later, once it's finished just publish with the `docs` option.

.github/workflows/npm-publish.yml
leordev: Oh as for the first question I think I answered above. Using `npm` when in a rush will just publish to npm and ignore the docs publishing...

Good question about the `always()`, I read about it and shamelessly copied from another template but it's for overriding the dependency imposed by `needs: tbdocs-publish` since that job won't run. (Let me do a test if it's really necessary or not).

leordev: yep, just confirmed on https://github.com/TBD54566975/tbdex-js/pull/38/commits/159e26de5d63c7cb08a86f8192114ed9f2235588 that the `always()` is needed.

Look at this run: https://github.com/TBD54566975/tbdex-js/actions/runs/6396868701

The `always` attribute forces the second job to run even if the first one was skipped.

.github/workflows/npm-publish.yml
leordev: Oh, I re-read your question and I just noticed you asked _where_ you would set these values! Sorry, it's right when you kickoff these manual release workflows, on the GitHub interface:
image

Manual workflows are enabled when you set the option `workflow_dispatch` attribute.

.github/workflows/npm-publish.yml
KendallWeihe: ah very interesting, all of the above. okay cool! I'm cool with merging ✅

PR #41, opened 2023-10-04
Filter tbdocs-reporter to only changed files

leordev:
Introduces the flag `report_changed_scope_only: true`, which will filter the tbdocs reporter to show only the modified lines by the PR.

Working evidence:
image


KendallWeihe:
Ah, nice! Much better!

KendallWeihe (approved):

PR #53, opened 2023-10-22
replace CBOR with JCS

mistermoe:
### `@tbdex/protocol`
Replaced CBOR with [JSON Canonicalization Scheme](https://datatracker.ietf.org/doc/html/rfc8785)

Using CBOR for deterministic serialization proved to be more difficult than originally anticipated specifically because of the lack of consistent implementations of the same CBOR sorting algorithm.

Related: https://github.com/TBD54566975/tbdex/pull/177

--- 

Refactored `Crypto`:
* Prior to this PR, `sign` allowed for either `detachedPayload` as a base64url string or `payload` as an object. If `detachedPayload` was provided, the resulting signature would be a detached JWS. If `payload` was provided, it would be JSON serialized and base64url encoded. Looking back, this felt awkward and confusing. Refactored such that `payload` is _always_ expected to be a `UInt8Array` and `detached` is now a boolean. 
* Refactored `verify` similar to ^
* Added a bit more detail to TSDoc

Generally speaking, `Crypto.sign`, and `Crypto.verify` still feel conflated and a bit janky. This is largely because they support tbdex signatures (compact detached JWS) _in addition to_ JWTs. Ideally, `Crypto` would be specific to tbdex related cryptography only with most of the core cryptographic functionality coming from `@web5/crypto`. The current state of both methods is largely a holdover until `@web5/crypto` is updated

### `@tbdex/http-client`

Updated request token generation to use refactored `Crypto` functions


decentralgabe:
curious why we need deterministic canonicalization as opposed to plain jwt?

phoebe-lew (approved):

phoebe-lew:
> curious why we need deterministic canonicalization as opposed to plain jwt?

@decentralgabe check out https://github.com/TBD54566975/tbdex/tree/main/specs/protocol#why-cbor or the updated explanation here https://github.com/TBD54566975/tbdex/blob/90e172adc4e3553c785aa00081a282ba6aa60519/specs/protocol/README.md#why-json-canonicalization-scheme-jcs

decentralgabe:
> > curious why we need deterministic canonicalization as opposed to plain jwt?
> 
> @decentralgabe check out https://github.com/TBD54566975/tbdex/tree/main/specs/protocol#why-cbor or the updated explanation here https://github.com/TBD54566975/tbdex/blob/90e172adc4e3553c785aa00081a282ba6aa60519/specs/protocol/README.md#why-json-canonicalization-scheme-jcs

thanks @phoebe-lew I'm familiar with JCS and what it does - but that doesn't exactly answer my question. JWTs don't enforce any canonicalization scheme in practice for a wide variety of applications, but don't encounter the problems you highlight. It does not really matter if two implementations produce differently ordered payloads, as long as the properties you care about are present and the signature checks out.

I would argue that outside of testing use cases (where property order may be useful) there is no reason to introduce the overhead of a canonicalization scheme. It is simpler and less error prone to sign and verify JSON (and its bytes) as-is than to worry about ordering.

mistermoe:
@decentralgabe 

> It does not really matter if two implementations produce differently ordered payloads, as long as the properties you care about are present and the signature checks out.

> what specifically about tbDEX messages makes them prone to these errors that warrants introducing a canonicalization scheme?

the [signature](https://github.com/TBD54566975/tbdex/tree/json-canonicalization-scheme/specs/protocol#signatures) of a tbDEX message or resource is a compact _detached_ JWS. the detached payload is a [digest](https://github.com/TBD54566975/tbdex/tree/json-canonicalization-scheme/specs/protocol#digests) of the message or resource. Verifying the signature necessitates producing the payload. producing the same payload necessitates a canonicalization scheme. Our rationale behind selecting detached JWS is [here](https://github.com/TBD54566975/tbdex/tree/json-canonicalization-scheme/specs/protocol#signatures)


simple test showcasing that different property order produces different digests:

```ts
import { sha256 } from '@noble/hashes/sha256'
import { Convert } from '@web5/common'
import { expect } from 'chai'

describe.only('canonicalization discovery', () => {
  it('different property order produces different digests', () => {
    const a = { 'a': 'b', 'b': 'a' }
    const aBytes = Convert.object(a).toUint8Array()
    const aDigest = sha256(aBytes)
    const aDigestB64url = Convert.uint8Array(aDigest).toBase64Url()

    console.log('a:', aDigestB64url)

    const b = {'b': 'a', 'a': 'b'}
    const bBytes = Convert.object(b).toUint8Array()
    const bDigest = sha256(bBytes)
    const bDigestB64url = Convert.uint8Array(bDigest).toBase64Url()

    console.log('b:', bDigestB64url)

    expect(aDigestB64url).to.not.equal(bDigestB64url)
  })
})
```

decentralgabe:
> the [signature](https://github.com/TBD54566975/tbdex/tree/json-canonicalization-scheme/specs/protocol#signatures) of a tbDEX message or resource is a compact _detached_ JWS. the detached payload is a [digest](https://github.com/TBD54566975/tbdex/tree/json-canonicalization-scheme/specs/protocol#digests) of the message or resource. Verifying the signature necessitates producing the payload. producing the same payload necessitates a canonicalization scheme.

The rationale states:

> By using a detached signature, the original payload remains unaltered and can be transmitted or stored in its native format without being embedded in the signature itself.

If unaltered, no problem. If reconstructing, problem

mistermoe:
concretely, we ran into integrity mismatches when we stored a resource or message in a database (specifically postgres using the json data type). On retrieval, we fetched a message from the db and sent it over the wire. the receiving side ran into an integrity mismatch which is what led us to realize that the cbor lib we were using wasn't deterministically encoding. switching to a deterministic cbor lib alleviated the mismatch.

_Then_ when we moved over to building out the kotlin sdk, we cycled through two different cbor libs ([jackson](https://github.com/FasterXML/jackson-dataformats-binary/tree/2.16/cbor) and [kotlinx.serialization.cbor](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor/)), both of which provided varied support in their implementations of deterministic encoding. Implementation variance is surfaced in the [CBOR RFC](https://datatracker.ietf.org/doc/html/rfc8949#name-deterministically-encoded-c). the JS lib we were using describes its recommendations [here](https://www.npmjs.com/package/cborg#deterministic-encoding-recommendations)

i'm sure Getting both libs to produce the same encoded values is possible, but it'll certainly be time consuming.

mistermoe:
`WebPKI.org`'s Java implementation also prescribes [its own "standard"](https://cyberphone.github.io/javaapi/org/webpki/cbor/package-summary.html#deterministic-encoding) for deterministic cbor encoding. figuring out whether its standard matches JS's `cborg` standard and so on and so forth for all other cbor libraries we encounter sort of defeated the purpose as to why i chose cbor to begin with which was the incorrect assumption that CBOR encoding is invariably deterministic (variable determinism feels like an oxymoron)

KendallWeihe:
So if I understand the use case correctly, the use case for canonicalization occurs when we need to reconstruct a signature from a piece of data (resource or message) _and_ expect the signature to remain constant across the original construction and any subsequent reconstructions.

So what use cases exist wherein we need to reconstruct signatures **and** expect the signature to remain constant across constructions? 

The latter part of that statement is crucial, because to @decentralgabe's point, we can always prove integrity regardless of if the signature is constant across different constructions.

decentralgabe:
we should use [CWT](https://www.rfc-editor.org/rfc/rfc8392.html), in the ideal. I'm not familiar with the intricacies of implementation differences across languages. I trust you that this is not as simple/straightforward as one might think. My gut instinct is we should spend the time to get this to work.

If we don't have the time right now - OK, but then yes, I'll go back to my earlier point (which @KendallWeihe highlights).

@mistermoe 
> concretely, we ran into integrity mismatches when we stored a resource or message in a database (specifically postgres using the json data type). On retrieval, we fetched a message from the db and sent it over the wire. the receiving side ran into an integrity mismatch which is what led us to realize that the cbor lib we were using wasn't deterministically encoding. switching to a deterministic cbor lib alleviated the mismatch.

This reads to me as a DB serialization/deserialization issue more than anything else. In the past I've mitigated such issues by storing the b64URL JWS value alongside the decoded JSON.

phoebe-lew:
> So what use cases exist wherein we need to reconstruct signatures and expect the signature to remain constant across constructions?

@KendallWeihe we reconstruct the payload every time a message is received/verified as we're using detached payload. The signature is over the `metadata` and `data` fields, hence requiring canonicalization.

KendallWeihe:
> @KendallWeihe we reconstruct the payload every time a message is received/verified as we're using detached payload. The signature is over the `metadata` and `data` fields, hence requiring canonicalization.

Ah right. Because it's not merely an auth token but also an integrity token. And our integrity check is a matter of reconstructing the same exact signature; which is to say, the signature is also an integrity hash. @decentralgabe are you claiming there are ways to execute integrity checks which are order-agnostic?

decentralgabe:
> > So what use cases exist wherein we need to reconstruct signatures and expect the signature to remain constant across constructions?
> 
> @KendallWeihe we reconstruct the payload every time a message is received/verified as we're using detached payload. The signature is over the `metadata` and `data` fields, hence requiring canonicalization.

is the detached payload not sent as `BASE64URL(JWS Payload)`? if so can it be so there's no need for a canonicalization scheme?

mistermoe:
@decentralgabe na the payload is just the message itself minus the signature

mistermoe:
going to go ahead and merge this and then create an issue to further explore using CBOR

cabo:
> _Then_ when we moved over to building out the kotlin sdk, we cycled through two different cbor libs ([jackson](https://github.com/FasterXML/jackson-dataformats-binary/tree/2.16/cbor) and [kotlinx.serialization.cbor](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-cbor/kotlinx.serialization.cbor/-cbor/)), both of which provided varied support in their implementations of deterministic encoding. Implementation variance is surfaced in the [CBOR RFC](https://datatracker.ietf.org/doc/html/rfc8949#name-deterministically-encoded-c). the JS lib we were using describes its recommendations [here](https://www.npmjs.com/package/cborg#deterministic-encoding-recommendations)

I haven't checked whether these libraries actually offer deterministic encoding.
If they do, did you report the bugs you found upstream?

PR #55, opened 2023-10-24
test vectors

mistermoe:
write initial tests using test vectors


mistermoe:
closing in favor of #129

PR #58, opened 2023-10-25
changing get exchanges filter to have id instead of exchangeId

jiyoontbd:
per current spec [here](https://github.com/TBD54566975/tbdex/tree/main/specs/http-api#list-exchanges)

### DIFF
--- a/packages/http-client/CHANGELOG.md
+++ b/packages/http-client/CHANGELOG.md
@@ -1,5 +1,15 @@
 # @tbdex/http-client
 
+## 0.18.0
+
+### Minor Changes
+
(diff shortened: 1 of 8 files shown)


packages/http-client/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # @tbdex/http-client
 
+## 0.18.0
+
+### Minor Changes
+
+- Get Exchanges filter field name change from exchangeId to id
```
KendallWeihe, line 7: technically this is a major change but we're still pre-1.0 version so just calling it out for the sake of sanity

KendallWeihe (approved):

PR #61, opened 2023-10-30
Automate GitHub releases

### LINKED ISSUE #60: Simplify the release process
Releasing in the current flow involves 5 steps:

1. merge PR with changeset into main.
1. CI opens versioning PR when changeset file is detected
1. merge PR
1. Create GitHub release
1. CI publishes to npm

With some automation in place, we want to simplify it to 3 steps:

1. merge PR with changeset into main.
1. CI opens versioning PR when changeset file is detected
1. Merge PR
    - CI automagically triggers:
      - Create GitHub release
      - Publish to NPM

leordev:
This PR automates the trigger to create a GH Release when a changeset PR is merged.

Fix #60 (see full context context).

(Also it removes the TBDocs from the release pipeline because now we are planning to release default Typedoc similar to the  DWN ones).

### DIFF
--- a/.github/workflows/create-gh-release.yml
+++ b/.github/workflows/create-gh-release.yml
@@ -1,0 +1,59 @@
+name: Create GH Release
+
+on:
+  pull_request:
+    types:
+      - closed
+    branches:
+      - main
+  workflow_dispatch:
+
+# Allow only one concurrent deployment,but do NOT cancel in-progress runs as
+# we want to allow these release deployments to complete.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  create-release:
+    # a gh release will be created in the following conditions:
+    # 1. if it is a manual trigger (workflow_dispatch)
+    # 2. if the workflow was triggered by a merged changeset PR
+    if: |
+      (github.event_name == 'workflow_dispatch' ||
+      (github.event.pull_request.merged == true &&
+        (github.event.pull_request.head.ref == 'changeset-release/main' ||
+        contains(github.event.pull_request.labels.*.name, 'automated-release'))))
+    name: Create GH Release
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+        with:
+          node-version: 20
+          registry-url: https://registry.npmjs.org/
+
+      - name: Check if GitHub repo package version is latest
+        run: |
+          # get any package version from the repo
+          cd packages/protocol
+
+          # Fetch the version in the GitHub repo's package.json file.
+          REPO_VERSION=$(node -p "require('./package.json').version")
+          echo "REPO_VERSION=$REPO_VERSION" >> $GITHUB_ENV
+          echo "Repo Version: $REPO_VERSION"
+        shell: bash
+
+      # TODO: uncomment before merging
+    #   - name: Create GitHub Release
+    #     uses: softprops/action-gh-release@v1
+    #     with:
+    #       name: v${{ env.REPO_VERSION }}
+    #       tag_name: v${{ env.REPO_VERSION }}
+    #       draft: true
+    #       prerelease: true
+    #       generate_release_notes: true
--- a/.github/workflows/create-gh-release.yml
+++ b/.github/workflows/create-gh-release.yml
@@ -51,8 +51,8 @@ jobs:
       - name: Create GitHub Release
         uses: softprops/action-gh-release@v1
         with:
-          name: v${{ env.REPO_VERSION }}
-          tag_name: v${{ env.REPO_VERSION }}
-          draft: true
-          prerelease: true
+          name: v${{ env.REPO_VERSION }}-test
+          tag_name: v${{ env.REPO_VERSION }}-test
+          draft: false
+          prerelease: false
           generate_release_notes: true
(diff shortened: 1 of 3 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -25,59 +16,7 @@ permissions:
   id-token: write # necessary for NPM provenance
 
 jobs:
-  tbdocs-publish:
```
KendallWeihe, line 28: What's the plan with this?

.github/workflows/create-gh-release.yml
```diff
@@ -0,0 +1,59 @@
+name: Create GH Release
+
+on:
+  pull_request:
+    types:
+      - closed
```
KendallWeihe, line 6: interesting terminology choice [by GitHub](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) here... to me "closed" is not the same as "merged"

.github/workflows/npm-publish.yml
leordev: 1. keep using the reporter for our regular PRs (it will keep us in check regarding the missing docs and breaking api changes)
2. automatically publish the vanilla html generated by typedoc (similarly to [dwn-sdk-js](https://tbd54566975.github.io/dwn-sdk-js/) one)
3. we will link from the developer.tbd.website to this one, instead of publishing it there.

This is fruit of a conversation that I had with @ALRubinger and @mistermoe.

We agreed that developers will prefer to consume the API reference in their own ecosystem instead of getting used to the TBD website theming, which would probably have limitations compared to their natural tools. Think of javadoc for java, docs.rs for rust, pkg.go.dev for go developers and so on... 

We started with js which has an ecosystem that most libs have their own website docs, so it was fine, but we are stepping back to a more standard approach, so that the website does not have embedded api references only for the js libs and the rest leaves outside... (cc @angiejones)

leordev:
@KendallWeihe can I go ahead and merge now?

.github/workflows/create-gh-release.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Set up Node.js
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+        with:
+          node-version: 20
+          registry-url: https://registry.npmjs.org/
+
+      - name: Check if GitHub repo package version is latest
+        run: |
+          # get any package version from the repo
+          cd packages/protocol
```
KendallWeihe, line 43:
```suggestion
          cd packages/protocol # assumption that all packages have the same version
```

This is a safe assumption for now, but it may change eventually

KendallWeihe:
@leordev I think we should add some material to the README which explains the way releases work, from the point of a developer opening a PR to the point where they can go to [npmjs.com](https://www.npmjs.com/package/@tbdex/protocol) and see the new version

I wrote the [Changesets](https://github.com/TBD54566975/tbdex-js#changesets) section, but it's slightly misleading in that it's descriptive rather than prescriptive. I like what I've written, but it's a little daunting at first glance, and we may be better off to move it somewhere different and have a simple `Release` section which consists of an ordered list of steps

1. Open PR
2. `changeset-bot` will automatically comment (https://github.com/TBD54566975/tbdex-js/pull/30#issuecomment-1732721942) on the PR with a reminder & recommendations for semver
3. Run `pnpm changeset` locally and push changes (`.changet/*.md`)
4. Merge PR into `main`
5. `.github/workflows/create-gh-release.yml` Will automatically create a [GitHub Release](https://github.com/TBD54566975/tbdex-js/releases)
6. `.github/workflows/npm-publish.yml` Will automatically publish to NPM 

Do I have it correct?

Have we had discussions around documenting release flows? Will release flows be standard across all implementations?

.github/workflows/create-gh-release.yml
leordev: Good call! When it changes, we will need to define how we are going to be tagging the GH releases too I think.

leordev:
@KendallWeihe 

> Do I have it correct?

Yes!

> I think we should add some material to the README which explains the way releases work, from the point of a developer opening a PR to the point where they can go to [npmjs.com](https://www.npmjs.com/package/@tbdex/protocol) and see the new version

Done! Check it out: https://github.com/TBD54566975/tbdex-js/pull/61/commits/10a74e1a8736d86f81f5b67fbafe4d1582e576b7#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

> Have we had discussions around documenting release flows? Will release flows be standard across all implementations?

Not yet, we do have this issue to capture that! And I agree that we should converge into a standard for all implementations. Although each language will have its own toolset.

@mistermoe @KendallWeihe about this step: `Run pnpm changeset locally and push changes (.changet/*.md)` -- should the developer ALWAYS do that for EVERY PR? Or only for PRs that they intend to release a new version? If the former, each merge to main will derive a new release. Is that the intent?

KendallWeihe:
I still didn't love my wording, specifically this bit, 

> Whereafter, a `pnpm changeset version` command must be executed.

...which makes it sound like the developer must execute this step.

I took the liberty of adjusting things a bit further https://github.com/TBD54566975/tbdex-js/pull/61/commits/098c1e48de68babacb46720b5bbb7bdb971bca42 @leordev what'd you think? Feel free to adjust/undo/whatevs.

KendallWeihe (approved):

.github/workflows/npm-publish.yml
```diff
@@ -4,15 +4,6 @@ on:
   release:
     types: [created]
   workflow_dispatch:
-    inputs:
```
jiyoontbd, line 7: could you give me some context on why this block has been removed?

leordev: Since there's no more docs publishing in the release process it does not make sense to have these options anymore!

PR #91, opened 2023-11-22
Handle http-server callback errors and pass offering to RFQ handler

phoebe-lew:


mistermoe (approved):

mistermoe:
bypassing branch protection because @diehuxx was here for implementation.

PR #97, opened 2023-11-23
Small updates

mistermoe:
- Added requester DID in filter passed to `ExchangesApi.getExchanges`. Prior to this change, there's no way to prevent returning exchanges that the requester wasn't a participant of
- Added `did:dht` resolution
- Fixed `DevTools.createJwt`


michaelneale (approved):
nice

phoebe-lew (approved):

PR #100, opened 2023-11-28
Refactor sign

mistermoe:
> [!IMPORTANT]
> This PR changes the Public API surface for both `@tbdex/http-client` and `@tbdex/protocol`

This PR introduces a change to the arguments passed to functions that sign things, namely:
* `Message.sign` (e.g. `rfq.sign`, `quote.sign` etc.)
* `Resource.sign` (e.g. `offering.sign`)

### Before
```typescript
const did = await DidDhtMethod.create()
const rfq = Rfq.create({
  metadata : { from: did.did, to: 'did:ex:pfi' },
  data     : rfqData
})

const { privateKeyJwk } = did.keySet.verificationMethodKeys[0]
let verificationMethodId = `${did.did}#${did.document.verificationMethod[0].id}`

await rfq.sign(privateKeyJwk, verificationMethodId)
```

### After
```typescript
const did = await DidDhtMethod.create()
const rfq = Rfq.create({
  metadata : { from: did.did, to: 'did:ex:pfi' },
  data     : rfqData
})

await rfq.sign(did)
```

## Rationale
* The proposed change prevents the caller from having to think or know about `privateKeyJwk` or verification method ids
* Aligns with the approach `tbdex-kt` and `tbdex-rs` have taken
* Sets us up for the key management changes coming to `web5-js`

### DIFF
--- a/.changeset/seven-bears-stare.md
+++ b/.changeset/seven-bears-stare.md
@@ -1,0 +1,7 @@
+---
+"@tbdex/http-client": minor
+"@tbdex/http-server": minor
+"@tbdex/protocol": minor
+---
+
+Refactored `sign` to take `PortableDid` as an argument instead of `privateKeyJwk` and `kid`
--- a/packages/http-client/src/client.ts
+++ b/packages/http-client/src/client.ts
@@ -1,5 +1,5 @@
 import type { DataResponse, ErrorDetail, ErrorResponse, HttpResponse } from './types.js'
-import type { PrivateKeyJwk as Web5PrivateKeyJwk } from '@web5/crypto'
+import type { PortableDid } from '@web5/dids'
 import type {
   ResourceMetadata,
   MessageModel,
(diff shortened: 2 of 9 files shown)


.changeset/seven-bears-stare.md
```diff
@@ -0,0 +1,7 @@
+---
+"@tbdex/http-client": minor
+"@tbdex/http-server": minor
+"@tbdex/protocol": minor
```
KendallWeihe, line 4: Technically these changes are major changes, as they break any existing app calls to `.sign()`

But, if we want to artificially keep the major version number lower since we're still under active dev then that's cool with me

KendallWeihe (approved):
love this so much ❤️

michaelneale (approved):
really nice change, I like it, but keep in mind should be more than a minor change in versions ideally as it does require some changes.

.changeset/seven-bears-stare.md
mistermoe: agreed! and yep just trying to keep the version number under 1.0.0 atm

mistermoe:
agreed @michaelneale ! been trying to keep the version under `1.0.0` until we're ready. been using minor versions as major until then. totally open to a different approach though

PR #110, opened 2023-12-05
Decouples security scanning analysis and add FOSSA

leordev:
- We are decoupling the security scanning to be able to run security checks on approved forks
- Plus, taking the opportunity to add FOSSA because we are assessing their security tooling and SBOMs generation

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -18,8 +18,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  security-action:
-    name: Security Analysis
+  snyk:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/security.yml
```diff
@@ -0,0 +1,62 @@
...
+
+on:
+  pull_request_target:
+    branches:
+      - main
+
+  push:
+    branches:
+      - main
+      - leordev/* # todo: remove this line before merging
```
KendallWeihe, line 11: 👀

KendallWeihe (approved):
So we're doing both Snyk and FOSSA? There is overlap there, yeah? But I presume they both offer their own distinct value-adds

leordev:
> So we're doing both Snyk and FOSSA? There is overlap there, yeah? But I presume they both offer their own distinct value-adds

Yes! They overlap but we are experimenting both to decide which one to adopt, or either move on with both for having double security layers. FOSSA provides SBOMs and Rust, Snyk provides advanced static analysis... So we are playing with both things!

mistermoe (approved):

PR #112, opened 2023-12-05
Replace sephereon/PEX presentation definitions with web5 js VCs implementation

kirahsapong:
This PR addresses #25

### DIFF
--- a/.changeset/few-donkeys-smile.md
+++ b/.changeset/few-donkeys-smile.md
@@ -1,0 +1,5 @@
+---
+"@tbdex/protocol": patch
+---
+
+Replace pex implementation with web5


KendallWeihe:
> explicitly types our offering's `get requiredClaims`

I believe the issue is because `PresentationDefinitionV2` is a type alias in `web5-js`; [see here](https://github.com/TBD54566975/web5-js/blob/acc002028dac186ad7e02f427dd92aed967f8df1/packages/credentials/src/presentation-exchange.ts#L12).

So I'm guessing the TypeScript compiler is replacing the type reference w/ `@sphereon/pex-models`, but that's not what is defined as the `OfferingData.requiredClaims` type [from here](https://github.com/TBD54566975/tbdex-js/blob/2a6b13410db8d451b6d862b09796c7a923222411/packages/protocol/src/types.ts#L88)

So yeah... it's an odd ball out, in that it's the only `get()` method with an explicit return type, but I guess that's okay.

The other option is, we could change `web5-js`'s `presentation-exchange.ts` to not do a type alias, and just directly export the sphereon type, so change this (how it currently is)...

```typescript
import type { PresentationDefinitionV2 as PexPresDefV2 } from '@sphereon/pex-models';

//...

export type PresentationDefinitionV2 = PexPresDefV2
```

to this...

```typescript
import type { PresentationDefinitionV2 } from '@sphereon/pex-models';
export type { PresentationDefinitionV2 } from '@sphereon/pex-models';
```

If we keep it as-is, then I would classify this as technical debt, albeit it's an extremely small amount of debt. The other option is, I think we should include a comment above the `get requiredClaims()` accessor explaining what's going on here (maybe with a link to this comment).

packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -77,9 +74,9 @@ export class Rfq extends Message<'rfq'> {
    * @throws if rfq's claims do not fulfill the offering's requirements
    */
   async verifyClaims(offering: Offering | ResourceModel<'offering'>) {
-    const { areRequiredCredentialsPresent } = pex.evaluateCredentials(offering.data.requiredClaims, this.claims)
+    const requiredCredentialsPresent = PresentationExchange.selectCredentials(this.claims, offering.data.requiredClaims)
```
KendallWeihe, line 77: This is a nit-pick, but variable name is slightly misleading IMO, could do either one of these...

```suggestion
const requiredCredentialsPresent = PresentationExchange.selectCredentials(this.claims, offering.data.requiredClaims).length > 0
```

And then update the `if ()` condition below like this

```typescript
    if (!requiredCredentialsPresent) {
      throw new Error(`claims do not fulfill the offering's requirements`)
    }
```

Or, could do...

```suggestion
    const credentials = PresentationExchange.selectCredentials(this.claims, offering.data.requiredClaims)
```

and then keep the `if ()` condition as-is (I think I prefer this way)

packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -77,9 +74,9 @@ export class Rfq extends Message<'rfq'> {
    * @throws if rfq's claims do not fulfill the offering's requirements
    */
   async verifyClaims(offering: Offering | ResourceModel<'offering'>) {
-    const { areRequiredCredentialsPresent } = pex.evaluateCredentials(offering.data.requiredClaims, this.claims)
+    const credentials = PresentationExchange.selectCredentials(this.claims, offering.data.requiredClaims)
```
jiyoontbd, line 77: does this method return credentials or error encapsulated in an object?

nitro-neal:
```
  /**
   * Selects credentials that satisfy a given presentation definition.
   *
   * @param {string[]} vcJwts The list of Verifiable Credentials to select from.
   * @param {PresentationDefinitionV2} presentationDefinition The Presentation Definition to match against.
   * @return {string[]} selectedVcJwts A list of Verifiable Credentials that satisfy the Presentation Definition.
   */
  public static selectCredentials(
    vcJwts: string[],
    presentationDefinition: PresentationDefinitionV2
  ): string[] {
    this.resetPex();
    const selectResults: SelectResults = this.pex.selectFrom(presentationDefinition, vcJwts);
    return selectResults.verifiableCredential as string[] ?? [];
  }
  
  ```
  
  it retuns the list of Verifiable Credentials that can satisfy the Presentation Definition.

mistermoe: actually might be helpful to return the error. this way we can surface it back up to the caller

mistermoe (approved):

packages/protocol/src/message-kinds/rfq.ts
KendallWeihe: I propose we use exceptions for error cases. Two benefits I can think of:

1. The return type is scoped to the semantic use case, which makes naming more cohesive and self-evident (ex. `List<Credentials>` as opposed to `SelectCredentialsResponse` (which would include a `credentials` and `errors` arrays))
2. By decoupling the error from the type, we have flexibility to introduce deeper call stacks abstracted behind the `selectCredentials` down the road, which could bubble-up exceptions, rather than having to then concern ourselves with passing errors up the callstack in return types

The counter argument is it makes the consumer have to use `try/catch` which can become verbose, but it isn't much more verbose than the subsequent `if (selectCredentialsResponse.errors?.length) { ...`

nitro-neal:
This is merged in now  - https://github.com/TBD54566975/web5-js/pull/336

KendallWeihe:
> Would it make sense to only verify those claims that satisfy the PD?

Yeah, I think so! @jiyoontbd can you think of an instance where we need to verify claims which are not defined in the Offering? That is, the Offering has N-number of claim requirements, and the RFQ has (N+M)-number of requirements, and we need to verify all N+M of them?

Issue #111, opened 2023-12-05
Refactor `http-client`


mistermoe:
Currently [http-client]() methods returns [`DataResponse`](https://github.com/TBD54566975/tbdex-js/blob/main/packages/http-client/src/types.ts#L38-L45) or [`ErrorResponse`](https://github.com/TBD54566975/tbdex-js/blob/main/packages/http-client/src/types.ts#L47-L54), both of which are [`HttpResponse`](https://github.com/TBD54566975/tbdex-js/blob/main/packages/http-client/src/types.ts#L29-L36)s. The Developer Experience is a bit awkward because

1. returning `status` and `headers` doesn't really provide any benefit
2. it's not entirely evident how callers should figure out whether the response received includes an error

We may want to consider something a bit more high level that:
1. drops inclusion of `headers` and `status` in what's returned
2. throw errors instead of returning them

for example, `getOfferings` would just return `Offering[]` on success or throw an error otherwise.

PR #113, opened 2023-12-07
Refactor http client

kirahsapong:
# Summary
This PR addresses #111 

It adds custom error types including `RequestError`, `ResponseError`, `ValidationError`, custom validation error types including `InvalidDidError` and `InvalidServiceEndpointError`. 

The `http-client` is refactored to throw a custom error type in case of a failure, and return data only in case of success.

This PR only introduces more narrow custom error types for `ValidationError`s. We can track adding more narrow custom error types for request and response errors [here](https://github.com/TBD54566975/tbdex-js/issues/117).

### DIFF
--- a/.changeset/four-mangos-swim.md
+++ b/.changeset/four-mangos-swim.md
@@ -2,4 +2,4 @@
 "@tbdex/http-client": minor
 ---
 
-Introduces custom errors types and throws instead of returns on failure
+Introduces custom errors types and breaking changes: functions now throw instead of return on failure


packages/http-client/src/client.ts
```diff
@@ -114,43 +115,44 @@ export class TbdexHttpClient {
...
+  static async getOfferings(opts: GetOfferingsOptions): Promise<Offering[]> {
     const { pfiDid , filter } = opts
 
-    const pfiServiceEndpoint = await TbdexHttpClient.getPfiServiceEndpoint(pfiDid)
+    let pfiServiceEndpoint
+
+    try {
+      pfiServiceEndpoint = await TbdexHttpClient.getPfiServiceEndpoint(pfiDid)
+    } catch(e) {
+      throw new RequestError({ message: e['message'], recipientDid: pfiDid, cause: e })
```
KendallWeihe, line 129: Rather than catching and throwing the exception here, would it be more wise to throw the exception from within the `TbdexHttpClient.getPfiServiceEndpoint` function? This is the beauty of using exceptions for error cases, rather than the return type, in that we can bubble-up the exceptions throughout n-number of callstack layers

mistermoe:
@kirahsapong we thinking about refactoring the rest of `http-client` in this PR or refactoring 1 method per PR?

KendallWeihe (approved):
🎉

packages/http-client/src/errors/validation-error.ts
mistermoe, line 1: nice!

packages/http-client/src/client.ts
```diff
@@ -27,7 +28,7 @@ export class TbdexHttpClient {
    * @throws if recipient DID resolution fails
    * @throws if recipient DID does not have a PFI service entry
    */
-  static async sendMessage<T extends MessageKind>(opts: SendMessageOptions<T>): Promise<HttpResponse | ErrorResponse> {
+  static async sendMessage<T extends MessageKind>(opts: SendMessageOptions<T>): Promise<HttpResponse> {
```
mistermoe, line 31: thoughts on just returning `Promise<void>` here?

mistermoe:
@kirahsapong as much as i want to use `msw`, i think it may be in our best interests to stub `fetch` instead. I've been working on updating how we run browser tests, but even with those changes, conditionally switching between the msw import for node vs. browser is going to be tricky. 

Did you have an approach in mind for conditionally switching between `import { setupWorker } from 'msw/browser'` and `import { setupServer } from 'msw/node'`? 

Otherwise, i think we can follow [this example](https://github.com/TBD54566975/web5-js/blob/main/packages/dids/tests/did-ion.spec.ts#L46-L75) from web5-js to mock fetch!

packages/http-client/src/client.ts
mistermoe: yup good idea!

packages/http-client/src/client.ts
```diff
@@ -45,20 +45,12 @@ export class TbdexHttpClient {
...
-      return { status, headers }
-    } else {
-      // TODO: figure out what happens if this fails. do we need to try/catch?
-      const responseBody: { errors: ErrorDetail[] } = await response.json()
-      return {
-        status  : response.status,
-        headers : response.headers,
-        errors  : responseBody.errors
-      }
+    if (!response.ok) {
```
phoebe-lew, line 51: so if this function returns nothing, that means the response was ok?

phoebe-lew (approved):
nice nice, can cut a new release 🚀

packages/http-client/src/client.ts
mistermoe: yee

mistermoe:
oh yeah! last thing to note would be that we probably should add a `changeset` to trigger a version bump. we should highlight that this is a breaking change and send the changelog over to devrel incase they've built any examples or written guides using `TbdexHttpClient`

PR #115, opened 2023-12-14
Update currency string representation

diehuxx:
Update currency amount representation and validation based on this spec change: https://github.com/TBD54566975/tbdex/pull/206

## Notes
I added an `await Message.validate(<message-object>` to each `#create` method. Since the updated test-vectors in the [spec change PR](https://github.com/TBD54566975/tbdex/pull/206) include regex for decimal strings, we can use `Message.validate()` to check that currency strings are in fact valid.

### DIFF
--- a/packages/http-server/src/fakes.ts
+++ b/packages/http-server/src/fakes.ts
@@ -10,9 +10,9 @@ const offering = Offering.create({
     },
     payoutCurrency: {
       currencyCode : 'BTC',
-      maxSubunits  : '99952611'
+      maxAmount    : '999526.11'
     },
-    payoutUnitsPerPayinUnit : '0.000038',
+    rate : '0.000038',
     payinMethods            : [{
       kind                   : 'DEBIT_CARD',
       requiredPaymentDetails : {
--- a/packages/http-server/src/fakes.ts
+++ b/packages/http-server/src/fakes.ts
@@ -12,7 +12,7 @@ const offering = await Offering.create({
       currencyCode : 'BTC',
       maxAmount    : '999526.11'
     },
-    rate         : '0.000038',
+    payoutUnitsPerPayinUnit         : '0.000038',
     payinMethods : [{
       kind                   : 'DEBIT_CARD',
       requiredPaymentDetails : {
--- a/packages/http-server/src/fakes.ts
+++ b/packages/http-server/src/fakes.ts
@@ -1,7 +1,7 @@
 import { MessageKindClass, Offering, Rfq, Quote, Order, OrderStatus, Close } from '@tbdex/protocol'
 import { OfferingsApi, ExchangesApi } from './main.js'
 
-const offering = Offering.create({
+const offering = await Offering.create({
   metadata : { from: 'did:ex:pfi' },
   data     : {
     description   : 'Selling BTC for USD',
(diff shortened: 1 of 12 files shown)


packages/protocol/src/resource-kinds/offering.ts
```diff
@@ -31,7 +31,9 @@ export class Offering extends Resource<'offering'> {
     }
 
     const message = { metadata, data: opts.data }
-    return new Offering(message)
+    const offering = new Offering(message)
+    await Resource.validate(offering)
```
phoebe-lew, line 35: `Resource.validate` is not async

packages/protocol/src/message-kinds/order-status.ts
```diff
@@ -32,7 +33,9 @@ export class OrderStatus extends Message<'orderstatus'> {
     }
 
     const message = { metadata, data: opts.data }
-    return new OrderStatus(message)
+    const orderStatus = new OrderStatus(message)
+    await Message.validate(order)
```
phoebe-lew, line 37: ditto not async

packages/protocol/src/message-kinds/order.ts
```diff
@@ -31,6 +31,8 @@ export class Order extends Message<'order'> {
     }
 
     const message = { metadata, data: {} }
-    return new Order(message)
+    const order = new Order(message)
+    await Message.validate(order)
```
phoebe-lew, line 35: ditto not async

packages/protocol/src/message-kinds/quote.ts
```diff
@@ -32,7 +32,9 @@ export class Quote extends Message<'quote'> {
     }
 
     const message = { metadata, data: opts.data }
-    return new Quote(message)
+    const quote = new Quote(message)
+    await Message.validate(quote)
```
phoebe-lew, line 36: ditto not async

packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -44,7 +44,9 @@ export class Rfq extends Message<'rfq'> {
     // TODO: hash `data.payoutMethod.paymentDetails` and set `private`
 
     const message = { metadata, data: opts.data }
-    return new Rfq(message)
+    const rfq = new Rfq(message)
+    await Message.validate(rfq)
```
phoebe-lew, line 48: ditto not async

packages/http-server/src/fakes.ts
```diff
@@ -10,10 +10,10 @@ const offering = Offering.create({
     },
     payoutCurrency: {
       currencyCode : 'BTC',
-      maxSubunits  : '99952611'
+      maxAmount    : '999526.11'
     },
-    payoutUnitsPerPayinUnit : '0.000038',
-    payinMethods            : [{
+    rate         : '0.000038',
```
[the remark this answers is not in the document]
diehuxx: It was suggested in the issue and I liked it. But I'll revert if it's controversial.

packages/protocol/src/message-kinds/close.ts
```diff
@@ -31,7 +31,9 @@ export class Close extends Message<'close'> {
     }
 
     const message = { metadata, data: opts.data }
-    return new Close(message)
+    const close = new Close(message)
+    await Message.validate(close)
```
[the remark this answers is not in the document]
diehuxx: Decided to remove `Message.validate` entirely since it's requires `signature` to be present even though the message is not signed at this point.

diehuxx:
Closing this PR because I forgot about (woops) and then reimplemented this change but more thoroughly in https://github.com/TBD54566975/tbdex-js/pull/134.

Issue #94, opened 2023-11-23
Finish implementing `Rfq.verifyOfferingRequirements`


mistermoe:
[todos here](https://github.com/TBD54566975/tbdex-js/blob/main/packages/protocol/src/message-kinds/rfq.ts#L58)

PR #120, opened 2023-12-19
Finish validating rfq against provided offering

kirahsapong:
This PR addresses #94

### DIFF
--- a/.changeset/soft-lizards-peel.md
+++ b/.changeset/soft-lizards-peel.md
@@ -1,0 +1,5 @@
+---
+"@tbdex/protocol": patch
+---
+
+Adds more checks to validate an RFQ against a provided Offering


packages/protocol/tests/rfq.spec.ts
```diff
@@ -158,6 +158,156 @@ describe('Rfq', () => {
...
+      }
+    })
+    it('throws an error if payinMethod kind cannot be validated against the provided offering\'s payinMethod kinds', async () => {
+      const rfq = Rfq.create({
+        ...rfqData,
+        data: {
+          ...rfqData.data,
+          payinMethod: {
+            ...rfqData.data.payinMethod,
+            kind: 'POKEMON'
```
jiyoontbd, line 240: 😃

jiyoontbd (approved):

packages/protocol/tests/rfq.spec.ts
```diff
@@ -158,6 +158,156 @@ describe('Rfq', () => {
     })
   })
 
+  describe('verifyOfferingRequirements', () => {
+    let offering
+    let rfqData
```
KendallWeihe, line 163:
```suggestion
    let offering: Offering
    let rfqData: RfqData
```

KendallWeihe: wait, might need to be nullable too, whatevs not a big deal

KendallWeihe: actually, could you reuse what's in `rfq.spec.ts`? https://github.com/TBD54566975/tbdex-js/blob/e31fa7ba252e67574897fab180246023c6920a08/packages/protocol/tests/rfq.spec.ts#L7

oh, now I see it's of type `Rfq` not `RfqData` (variable name should just be `rfq` then)

KendallWeihe: (deleted a comment b/c I spoke too soon)

packages/protocol/tests/rfq.spec.ts
```diff
@@ -158,6 +158,156 @@ describe('Rfq', () => {
...
+    before(() => {
+      async function initMocks() {
```
KendallWeihe, line 165: I think you can make the entire `before` function `async` and then you don't need a `initMocks()` function block, everything could just go within the `before()` function block, like this...

```typescript
before(async () => {
```

KendallWeihe (approved):
lgtm, though the doc commenter is complaining about missing member variables?

Screenshot 2023-12-19 at 10 56 15 AM

PR #134, opened 2024-01-12
Update currency representation to decimal string

diehuxx:
Addresses https://github.com/TBD54566975/tbdex/pull/206

Also, in adding the decimalString regex to the JSON schema, I found a few tangentially related tests that should have failed but didn't. I've tweaked the tests to prevent leakages, and added additional validation of `data` properties upon Message/Resource creation so we can get the sweet regex on the decimal strings.

I added a somewhat jank util to help generate updated test vectors as we make spec changes. Try `pnpm run generate-test-vector`.


phoebe-lew (approved):

packages/protocol/package.json
```diff
@@ -55,6 +55,7 @@
     "@web5/crypto": "0.2.2",
     "@web5/dids": "0.2.2",
     "ajv": "8.12.0",
+    "bignumber.js": "^9.1.2",
```
jiyoontbd, line 58: curious what this import is used for?

packages/protocol/src/dev-tools.ts
```diff
@@ -26,6 +27,10 @@ export type RfqOptions = {
    * of the offering returned by {@link DevTools.createOffering}
    */
   sender: PortableDid
+  /**
+   * {@link @web5/dids#PortableDid} of the rfq receiver.
+   */
+  receiver?: PortableDid
```
jiyoontbd, line 33: what is this new field used for?

packages/protocol/src/dev-tools.ts
```diff
@@ -75,16 +80,25 @@ export class DevTools {
...
+  /**
+   * creates an example OfferingData. Useful for testing purposes
+   */
+  static createOfferingData(): OfferingData {
+    return {
       description   : 'Selling BTC for USD',
       payinCurrency : {
-        currencyCode : 'USD',
-        maxSubunits  : '99999999'
+        currencyCode: 'USD'
```
jiyoontbd, line 97: could we add in the `maxAmount`: '99999999.00` since it was in the original offeringdata?

jiyoontbd: maybe add `minAmount` for funsies too

packages/protocol/src/dev-tools.ts
```diff
@@ -135,37 +149,49 @@ export class DevTools {
         }
       }],
       requiredClaims: {
-        id     : '7ce4004c-3c38-4853-968b-e411bafcd945',
-        format : {
```
jiyoontbd, line 139: do we still need this format object or nah

packages/protocol/package.json
jiyoontbd: found it :)

packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -63,8 +65,13 @@ export class Rfq extends Message<'rfq'> {
       throw new Error(`offering id mismatch. (rfq) ${this.offeringId} !== ${offering.metadata.id} (offering)`)
     }
 
-    if (this.payinSubunits > offering.data.payinCurrency.maxSubunits) {
-      throw new Error(`rfq payinSubunits exceeds offering's maxSubunits. (rfq) ${this.payinSubunits} > ${offering.data.payinCurrency.maxSubunits} (offering)`)
+    if (offering.data.payinCurrency.maxAmount) {
```
jiyoontbd, line 68: should we add a check for minAmount too?

jiyoontbd (approved):

packages/protocol/src/dev-tools.ts
diehuxx: It's used to set the `to` property, aka the intended receiver of the RFQ.

packages/protocol/src/dev-tools.ts
diehuxx: `requiredClaims` is of type `PresentationDefinitionV2` which does take an optional `format` param. That type comes from an external library. This fixture doesn't need `format` for any tests that this is involved in. `createOfferingData()` and other functions in `dev-tools.ts` are mainly meant as basic fixures that can be used conveniently when testing, so they don't necessarily fill out all possible fields.

packages/protocol/src/dev-tools.ts
diehuxx: Updated

packages/protocol/package.json
diehuxx: For anyone else reviewing, `bignumber.js` is a library that does arithmetic on strings containing numbers in javascript, because javascript is bad at arithmetic on its own.

PR #139, opened 2024-01-16
make requiredPaymentDetails optional

### LINKED ISSUE #90: make `requiredPaymentDetails` optional
[this](https://github.com/TBD54566975/tbdex-js/blob/main/packages/protocol/src/types.ts#L112) should be optional as per the [spec](https://github.com/TBD54566975/tbdex/tree/main/specs/protocol#paymentmethod)

finn-tbd:
fixes #90


diehuxx (approved):

finn-tbd:
left this in draft state because I wasn't sure if further checks were needed. It looks like this field is used in a `protoc/src/message-kinds/rfq.ts` where it is assumed to be set. Should we have some safety around checking there? I'm missing a lot of context around what this field is and how it may be used.

diehuxx:
> It looks like this field is used in a protoc/src/message-kinds/rfq.ts where it is assumed to be set.

@finn-tbd Good catch. The fix for that would conflict with #141 (which will also make the TS compiler strict enough to catch errors like that in the future). Once #141 is merged, I can push a commit to this PR fixing `protoc/src/message-kinds/rfq.ts` is that's alright with you.

finn-tbd:
sounds good

diehuxx:
TODO:
- [ ] After https://github.com/TBD54566975/tbdex/pull/228 is merged, update tbdex submodule back to `main`.

phoebe-lew (approved):

finn-tbd:
Blocked pending https://github.com/TBD54566975/tbdex/pull/227

packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -146,28 +145,40 @@ export class Rfq extends Message {
...
     for (const paymentMethodMatch of paymentMethodMatches) {
+      if (!paymentMethodMatch.requiredPaymentDetails) {
+        // If requiredPaymentDetails is omitted, and paymentDetails is also omitted, we have a match
+        if (!rfqPaymentMethod.paymentDetails) {
+          return
+        }
+
+        // paymentDetails is present even though requiredPaymentDetails is omitted. This is unsatisfactory.
+        invalidPaymentDetailsErrors.add(new Error('paymentDetails must be omitted when requiredPaymentDetails is omitted'))
+        continue
```
KendallWeihe, line 166: rather than a `continue` here, could we do an `else` block?

KendallWeihe (approved):
one comment added, but not strictly required (opinionated take)

feel free to merge once this blocking PR is merged https://github.com/TBD54566975/tbdex/pull/227

packages/protocol/src/message-kinds/rfq.ts
diehuxx: works for me

PR #141, opened 2024-01-17
Stricten typescript in protocol package

diehuxx:
The other packages will be strictened in a follow up PR.

### DIFF
--- a/packages/protocol/src/crypto.ts
+++ b/packages/protocol/src/crypto.ts
@@ -2,7 +2,9 @@ import type {
   PrivateKeyJwk as Web5PrivateKeyJwk,
   CryptoAlgorithm,
   Web5Crypto,
-  JwsHeaderParams
+  JwsHeaderParams,
+  PrivateKeyJwk,
+  PublicKeyJwk
 } from '@web5/crypto'
 
 import { sha256 } from '@noble/hashes/sha256'
@@ -42,7 +44,7 @@ export type VerifyOptions = {
  */
 type SignerValue<T extends Web5Crypto.Algorithm> = {
   signer: CryptoAlgorithm,
-  options?: T,
+  options: T,
   alg: JwsHeader['alg'],
   crv: JsonWebKey['crv']
 }
@@ -99,18 +101,18 @@ export class Crypto {
   static async sign(opts: SignOptions) {
     const { did, payload, detached } = opts
 
-    const { privateKeyJwk } = did.keySet.verificationMethodKeys[0]
+    const privateKeyJwk = did.keySet.verificationMethodKeys?.[0]?.privateKeyJwk
 
-    const algorithmName = privateKeyJwk['alg'] || ''
-    const namedCurve = privateKeyJwk['crv'] || ''
+    const algorithmName = privateKeyJwk?.['alg'] || ''
+    let namedCurve = Crypto.extractNamedCurve(privateKeyJwk)
     const algorithmId = `${algorithmName}:${namedCurve}`
 
     const algorithm = this.algorithms[algorithmId]
     if (!algorithm) {
-      throw new Error(`${algorithmId} not supported`)
+      throw new Error(`Algorithm (${algorithmId}) not supported`)
     }
 
-    let verificationMethodId = did.document.verificationMethod[0].id
+    let verificationMethodId = did.document.verificationMethod?.[0]?.id || ''
     if (verificationMethodId.startsWith('#')) {
       verificationMethodId = `${did.did}#${verificationMethodId}`
     }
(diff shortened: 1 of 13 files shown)


packages/protocol/src/did-resolver.ts
```diff
@@ -47,7 +48,7 @@ export type DidResource = DidDocument | VerificationMethod | DidService
  * @throws if DID cannot be resolved
  * @beta
  */
-export async function deferenceDidUrl(didUrl: string): Promise<DidResource> {
+export async function deferenceDidUrl(didUrl: string): Promise<DidResource | undefined> {
```
diehuxx, line 51: How should `deferenceDidUrl()` behave if verification methods or services are missing? Should we throw or is returning undefined sufficient?

packages/protocol/src/message.ts
```diff
@@ -16,7 +16,7 @@ import { PortableDid } from '@web5/dids'
 export abstract class Message<T extends MessageKind> {
   private _metadata: MessageMetadata<T>
   private _data: MessageKindModel<T>
-  private _signature: string
+  private _signature: string | undefined
```
diehuxx, line 19: Messages are signed after creation, so they might not yet have a signature

packages/protocol/src/resource.ts
```diff
@@ -15,7 +15,7 @@ import { PortableDid } from '@web5/dids'
 export abstract class Resource<T extends ResourceKind> {
   private _metadata: ResourceMetadata<T>
   private _data: ResourceKindModel<T>
-  private _signature: string
+  private _signature: string | undefined
```
diehuxx, line 18: Resources are signed after creation, so they might not yet have a signature

packages/protocol/src/types.ts
```diff
@@ -121,8 +121,10 @@ export type MessageModel<T extends MessageKind> = {
   metadata: MessageMetadata<T>
   /** The actual message content */
   data: MessageKindModel<T>
+  /** Private data that must not be in the main  */
+  private?: T extends 'rfq' ? Record<string, any> : never
   /** signature that verifies that authenticity and integrity of a message */
-  signature: string
+  signature?: string
```
diehuxx, line 127: Instead of using Typescript to guarantee the presence of a signature, we should be (and are) using JSON schemas for validation.

packages/protocol/tests/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "strict": true,
+    "useUnknownInCatchVariables": false,
```
diehuxx, line 5: In tests, I think it's fine to be lax about this rule

jiyoontbd (approved):
looks great!

phoebe-lew (approved):

packages/protocol/src/crypto.ts
```diff
@@ -42,7 +44,7 @@ export type VerifyOptions = {
  */
 type SignerValue<T extends Web5Crypto.Algorithm> = {
   signer: CryptoAlgorithm,
-  options?: T,
+  options: T,
```
phoebe-lew, line 47: is this actually not nullable?

packages/protocol/src/types.ts
phoebe-lew: This does seem to contradict the spec, which says that this field is required?

diehuxx: https://github.com/TBD54566975/tbdex-js/pull/141#issuecomment-1899080621

packages/protocol/src/did-resolver.ts
diehuxx: @andresuribe87 Any thoughts here?

packages/protocol/src/crypto.ts
diehuxx: I'm not sure what the right approach is. It looks like all the `SignerValue`s have `options` present. And it's not clear what to do in `Crypto.sign` or `Crypto.verify` if `options === undefined`. Currently, if `options === undefined`, we would have bug when it is passed to `Web5` `CryptoAlgorithm#sign` or `CryptoAlgorithm#verify`.

I'm open to any opinions. @frankhinek If you have more context on web5 crypto, what approach would you take here?

packages/protocol/src/did-resolver.ts
andresuribe87: Given that DID URL dereferencing is defined in https://www.w3.org/TR/did-core/#did-url-dereferencing, I would suggest that this function should return something similar to what's defined in the spec. 

If that's the case, then I can see the case for returning an object similar to the example in https://www.w3.org/TR/did-core/#did-url-dereferencing. In particular, for the example mentioned, the property named `didUrlDereferencingMetadata`, would have an object which contains an `error` property. The value of the property would be `notFound`.

diehuxx: Ah, looks like web5/dids has an implementation of DID URL dereferencing but it's not in the latest version. I'll just mark a TODO #147

PR #142, opened 2024-01-18
adding `replyTo` fields inside RfqData.

jiyoontbd:


packages/http-client/src/client.ts
```diff
@@ -28,7 +28,8 @@ export class TbdexHttpClient {
    * @throws if recipient DID does not have a PFI service entry
    */
   static async sendMessage<T extends MessageKind>(opts: SendMessageOptions<T>): Promise<void> {
-    const { message } = opts
+    const { message, replyTo } = opts
```
jiyoontbd, line 31: i know SendMessageOptions limits the replyTo to be a string only when message is of Rfq type, but i think it would be nice to check that in the code and not just rely on the type system. not sure how to do this using typescript type voodoo tho

diehuxx: If you want a runtime check, you can do that with regular js using `typeof replyTo === 'string'`

diehuxx: oh wait looks like you've got a similar check below where you check if it's undefined. that should suffice

packages/http-client/tests/client.spec.ts
```diff
@@ -20,19 +20,19 @@ sinon.stub(Message, 'verify').resolves('123')
 describe('client', () => {
   beforeEach(() => getPfiServiceEndpointStub.resolves('https://localhost:9000'))
 
-  describe('sendMessage', async () => {
-    let mockMessage: Rfq
+  describe.only('sendMessage', async () => {
```
diehuxx, line 23:
```suggestion
  describe('sendMessage', async () => {
```

jiyoontbd: removed from this and also another `it.only` i used to run the single test i wrote :)

packages/http-client/tests/client.spec.ts
```diff
@@ -53,29 +53,67 @@ describe('client', () => {
...
+        json : () => Promise.resolve()
+      } as Response)
+
+      try {
+        await TbdexHttpClient.sendMessage({message: mockRfqMessage})
+      } catch (e) {
+        expect.fail()
+      }
+    })
+    it.only('should throw errors when sending Order with replyTo field', async () => {
```
diehuxx:
```suggestion
    it('should throw errors when sending Order with replyTo field', async () => {
```

packages/http-client/src/client.ts
```diff
@@ -37,12 +38,16 @@ export class TbdexHttpClient {
     const pfiServiceEndpoint = await TbdexHttpClient.getPfiServiceEndpoint(pfiDid)
     const apiRoute = `${pfiServiceEndpoint}/exchanges/${exchangeId}/${kind}`
 
+    if (jsonMessage.metadata.kind != 'rfq' && replyTo) {
+      throw new RequestError({ message: `Failed to send message to ${pfiDid}. replyTo field should not be present in a ${jsonMessage.metadata.kind} message.`, recipientDid: pfiDid, url: apiRoute })
```
diehuxx, line 42: I'm kinda overthinking this so bear with me. The type system already checks this for us, so in typescript-land we don't need this `if`. In vanilla JS, we shouldn't really need it either since for messages other than RFQ this method  should only ignore `replyTo` anyway.

packages/http-client/src/client.ts
```diff
@@ -277,6 +282,7 @@ export class TbdexHttpClient {
 export type SendMessageOptions<T extends MessageKind> = {
   /** the message you want to send */
   message: Message<T> | MessageModel<T>
+  replyTo?: T extends 'rfq' ? string : never
```
diehuxx, line 285: Doc comment pls

packages/http-client/src/client.ts
jiyoontbd: yea i was thinking about this for a bit as well!

i think it is technically possible to write something like this and have it compile in ts - as i was able to write a test like this. but if you think it's unnecessary as we are sane and civilized and thus use types

```js
let mockOrderMessage // no type! if i type this to Order then the sendMessage({..}) compiler yells at me
      try {
        mockOrderMessage = Order.create({
          metadata: {
            from       : dhtDid.did,
            to         : dhtDid.did,
            exchangeId : 'rfq_123'
          }
        })
        await TbdexHttpClient.sendMessage({message: mockOrderMessage, replyTo: 'https://tbdex.io/callback'})
        expect.fail()
      } catch (e) { 
      // asserts the exception thrown
```

diehuxx (approved):

packages/http-client/src/client.ts
jiyoontbd: i decided to trust typescript

phoebe-lew (approved):

PR #146, opened 2024-01-19
Setup CodeQL SAST

leordev:


diehuxx (approved):

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,86 @@
...
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: '33 1 * * 5'
```
jiyoontbd, line 20: is there a reason why the cron schedule is different here compared to the one in your tbdex-kt PR?

jiyoontbd (approved):

.github/workflows/codeql.yml
michaelneale: how come it is on a schedule at all vs pull requests?

leordev: > is there a reason why the cron schedule is different here compared to the one in your tbdex-kt PR?
It's randomly generated by GH when Enabling the CodeQL scan. I don't think the schedule time actually matter as long as we have the scan being executed weekly. (I'm going to move forward with the merge but we can revisit the weekly scanning time later if we want to have it running at the same time for all our repos).

> how come it is on a schedule at all vs pull requests?
It runs both on PRs, main pushes AND on a schedule. The reason for running on a schedule is that new vulns are discovered every day. So, even after merging something, we need to be scanning the codebase in a consistent basis.

PR #126, opened 2023-12-22
[http-client] Refactor request token to be a JWT

mistermoe:
addresses #122

Issue #122, opened 2023-12-20
[http-client] refactor request token to be a JWT


mistermoe:
context [here](https://github.com/TBD54566975/tbdex/issues/210).

* refactor [`TbdexHttpClient.generateRequestToken`](https://github.com/TBD54566975/tbdex-js/blob/main/packages/http-client/src/client.ts#L268-L281) to produce a JWT.  can make use of the [`JWT`](https://github.com/TBD54566975/web5-js/blob/main/packages/credentials/src/jwt.ts) utility class we just added to `@web5/credentials`.
* refactor [`TbdexHttpClient.verify`](https://github.com/TBD54566975/tbdex/issues/210) to verify JWT. probably should change the method name to be a bit more explicit too. maybe `verifyRequestToken`

PR #148, opened 2024-01-20
adding default typescript version to align on

jiyoontbd:
# Summary

Aligning on typescript version to be at `5.2.2` for VS Code `settings.json` and top level `package.json`

Part of an effort to break apart [this PR](https://github.com/TBD54566975/tbdex-js/pull/126) into smaller ones. Related to #122


diehuxx (approved):

michaelneale (approved):
I expect this is a very good idea to save confusion

PR #149, opened 2024-01-20
bumping web5 versions

### LINKED ISSUE #147: Replace bespoke DID dereference implementation with web5/dids after update
Context: https://github.com/TBD54566975/tbdex-js/pull/141#discussion_r1454398823

jiyoontbd:
# Summary
Upgrading to latest versions of `@web5` packages (dids, common, crypto, credentials)
Removed dependency from `Jose` library in `Crypto.ts`
Using `DidResolver.dereference()` from `@web5/dids` package, which closes #147 
Using `VerifiableCredentials.create()` from `@web5/credentials` package instead of `DevTools.createCredential()`
Part of an effort to break apart https://github.com/TBD54566975/tbdex-js/pull/126 into smaller ones. Related to #122

### DIFF
--- a/packages/http-client/package.json
+++ b/packages/http-client/package.json
@@ -50,17 +50,21 @@
   },
   "dependencies": {
     "@tbdex/protocol": "workspace:*",
-    "@web5/common": "0.2.1",
-    "@web5/crypto": "0.2.2",
-    "@web5/dids": "0.2.2",
-    "query-string": "8.1.0"
+    "@web5/common": "0.2.2",
+    "@web5/credentials": "0.4.1",
+    "@web5/crypto": "0.2.4",
+    "@web5/dids": "0.2.4",
+    "ms": "2.1.3",
+    "query-string": "8.1.0",
+    "typeid-js": "0.3.0"
   },
   "devDependencies": {
     "@playwright/test": "1.34.3",
     "@types/chai": "4.3.5",
     "@types/eslint": "8.37.0",
     "@types/mocha": "10.0.1",
-    "@types/sinon": "^17.0.2",
+    "@types/ms": "0.7.34",
+    "@types/sinon": "17.0.2",
     "@typescript-eslint/eslint-plugin": "5.59.0",
     "@typescript-eslint/parser": "5.59.0",
     "chai": "4.3.10",
@@ -78,7 +82,7 @@
     "mkdirp": "3.0.1",
     "mocha": "10.2.0",
     "rimraf": "4.4.0",
-    "sinon": "15.0.2",
+    "sinon": "17.0.1",
     "typedoc": "0.25.0",
     "typedoc-plugin-markdown": "3.16.0",
     "typescript": "5.2.2"
--- a/packages/http-client/package.json
+++ b/packages/http-client/package.json
@@ -64,7 +64,7 @@
     "@types/eslint": "8.37.0",
     "@types/mocha": "10.0.1",
     "@types/ms": "0.7.34",
-    "@types/sinon": "17.0.2",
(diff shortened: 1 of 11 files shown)


packages/http-client/package.json
```diff
@@ -50,17 +50,21 @@
...
-    "@web5/crypto": "0.2.2",
-    "@web5/dids": "0.2.2",
-    "query-string": "8.1.0"
+    "@web5/common": "0.2.2",
+    "@web5/credentials": "0.4.1",
+    "@web5/crypto": "0.2.4",
+    "@web5/dids": "0.2.4",
+    "ms": "2.1.3",
+    "query-string": "8.1.0",
+    "typeid-js": "0.3.0"
```
diehuxx, line 59: Why are we adding `ms` and `typeid-js` to `http-client` in this PR? If `ms` is necessary in `dependencies`, why is `@types/ms` in `devDepencencies`?

packages/http-client/package.json
```diff
@@ -78,7 +82,7 @@
     "mkdirp": "3.0.1",
     "mocha": "10.2.0",
     "rimraf": "4.4.0",
-    "sinon": "15.0.2",
+    "sinon": "17.0.1",
```
diehuxx, line 85: Should `sinon` and `@types/sinon` have matching versions? Only different by patch version so probably not a big deal regardless

packages/protocol/src/crypto.ts
```diff
@@ -101,32 +102,30 @@ export class Crypto {
   static async sign(opts: SignOptions) {
     const { did, payload, detached } = opts
 
-    const privateKeyJwk = did.keySet.verificationMethodKeys?.[0]?.privateKeyJwk
+    const privateKeyJwk = did.keySet.verificationMethodKeys[0].privateKeyJwk as JwkParamsEcPrivate | JwkParamsOkpPrivate
 
-    const algorithmName = privateKeyJwk?.['alg'] || ''
-    let namedCurve = Crypto.extractNamedCurve(privateKeyJwk)
+    const algorithmName = privateKeyJwk['alg'] || ''
+    const namedCurve = privateKeyJwk['crv'] || ''
```
diehuxx, line 108: CI is failing because you're reverting some of the changes from #141. Might need to rebase.

diehuxx (commented):
Updating web5 deps unblocks #147, which has a TODO in `protocol/src/did-resolver.ts`. We can tackle that in this PR or a follow up, up to you @jiyoontbd.

packages/http-client/package.json
jiyoontbd: good catch, ty!

jiyoontbd:
ci is still failing because of test vector tests in packages/protocol, but that will be addressed in this [test vector specific PR](https://github.com/TBD54566975/tbdex-js/pull/152)

diehuxx (approved):

packages/http-client/package.json
mistermoe: yee that's correct @jiyoontbd

jiyoontbd:
> 🎉 nice!
> 
> Can we also clean up [this TODO](https://github.com/TBD54566975/tbdex-js/blob/de22b4fead19c3e38d91c47b56abb94e98f47de6/packages/protocol/src/resource-kinds/offering.ts#L68-L70) in `offering.ts` which is just removing the type annotation following the web5/credentials pkg update?
> 
> ```
> - get requiredClaims(): PresentationDefinitionV2 { ... }
> + get requiredClaims() { ... }
> ```
> 
> I can push a quick commit for that if you prefer - lmk!

@kirahsapong removed the return type

PR #156, opened 2024-02-03
Simplify types, inheritance structure, and API

diehuxx:
## All Changes
* Simplify inheritance structure of `Message` and its subclasses and `Resource` and its subclasses.
* Simplify types in `protocol` and `http-server` packages. Remove generics wherever possible. Replace generics with intersection types which are easier to read. Remove all types which could be considered "typescript neckbeardery"
* Remove `Message.factory` and `Resource.factory`. Introduce standalone functions `parseMessage` and `parseResource`, reducing coupling and circular dependency.
* Improve test coverage tests for `Message` and `Rfq`. In follow up PRs, I will further improve test coverage in the all three packages. I'm not testing in this PR because it's already a big diff.
* Slight refactor to the API of `TbdexHttpServer` for adding callback functions. http-server has no tests currently, so tests for this new API will come in a follow up PR.

### tbdex/protocol

I updated the README to reflect the new uses.

All uses of `MessageKindClass` and `ResourceKindClass` have been replaced with `Message` and `Resource`.

`Message.parse` and `Resource.parse` have been replaced with `parseMessage()` and `parseResource()`.

Each message-kind subclass and resource-kind subclass (e.g. `Rfq`, `Offering`, etc.) now have their own dedicated `.parse()` method which checks that the input is the expected message kind. For example, `Rfq.parse(inputJson)` will return an instance of the `Rfq` class if and only if `inputJson` is a valid, signed RFQ.

`Message.verify()`, `Resource.verify()`, `Message.validate()`, and `Resource.validate()` are no longer static methods. They have been replaced with instance methods

### tbdex/http-server
Updated to reflect type changes from `tbdex/protocol`, most notably `MessageKindClass` -> `Message`.

Alter API for adding callbacks methods to Tbdex server. Removes methods `get` and `submit`, replacing them with
* `onSubmitRfq(callback: SubmitRfqCallback): void`
* `onSubmitOrder(callback: SubmitOrderCallback): void`
* `onSubmitClose(callback: SubmitCloseCallback): void`
* `onGetExchanges(callback: GetExchangesCallback): void`
* `onGetOfferings(callback: GetOfferingsCallback): void`


mistermoe:
diving into this now

packages/protocol/tests/offering.spec.ts
```diff
@@ -1,64 +1,81 @@
-import { Offering } from '../src/main.js'
+import { Offering, OfferingData, parseResource } from '../src/main.js'
```
mistermoe, line 1: `OfferingData` is unused

packages/protocol/tests/offering.spec.ts
mistermoe, line 1: I think there's some formatting issues in this file that the linter can auto fix

mistermoe:
## regarding changes made to `@tbdex/protocol`

Definitely feels like an overall improvement! shaved the neckbeard at the cost of what looks like repetition on initial glance but i think could be argued isn't actually. Either way, the cognitive overhead required to add a new message kind or make a change is _significantly_ reduced which is awesome! Just compared adding a new message kind on `main` and then on your branch and it's significantly easier here. 

Also, it doesn't seem like we sacrificed any type specificity at the public api layer. VSCode is still correctly guiding me through creating all types of messages. Biggest wins imo are: 
* ditching `MessageKind` which was super annoying because it was difficult to decide whether to `Message` or `MessageKind`. 
* ditching the hack i added to prevent circular dependencies

## regarding changes made to `@tbdex/http-server`:
Still forming an opinion here because the api surface did notably change.

## Other comments:
* Can you include a writeup of all the breaking changes at the public api level with examples so that we can share it with devrel? i think they have guides that are using what's currently on main and we'll need a way to convey what's changed and how.

packages/protocol/tests/offering.spec.ts
diehuxx: fixed

packages/protocol/tests/offering.spec.ts
diehuxx: linted

diehuxx:
> Definitely feels like an overall improvement! shaved the neckbeard at the cost of what looks like repetition on initial glance but i think could be argued isn't actually.

Glad you think so! I made a deliberate choice to allow small amounts of repetition in order to make the API surface and the implementation simpler.

> Can you include a writeup of all the breaking changes at the public api level with examples so that we can share it with devrel? i think they have guides that are using what's currently on main and we'll need a way to convey what's changed and how.

Absolutely. I'll add details to the PR description tomorrow.

packages/http-server/src/http-server.ts
```diff
@@ -100,21 +95,43 @@ export class TbdexHttpServer {
...
+   */
+  onSubmitClose(callback: SubmitCloseCallback): void {
+    this.callbacks['close'] = callback
+  }
+
+  /**
+   * Set up the callback for the GetExchanges endpoint
+   * @param callback
    */
-  submit<T extends SubmitKind>(messageKind: T, callback: SubmitCallbacks[T]) {
```
jiyoontbd, line 107: i think we lost this `submit()` method that allows the consumer of this http-server library from writing callbacks like `server.submit<Order>(...) { // biz logic for processing an Order }` ?

jiyoontbd: oh, nvm you just broke it out into message-specific ones.

packages/protocol/package.json
```diff
@@ -82,7 +83,7 @@
     "build:esm": "rimraf dist/esm dist/types && tsc",
     "build:cjs": "rimraf dist/cjs && tsc -p tsconfig.cjs.json && echo '{\"type\": \"commonjs\"}' > ./dist/cjs/package.json",
     "build:browser": "rimraf dist/browser.mjs dist/browser.js && node build/bundles.js",
-    "test:node": "rimraf tests/compiled && pnpm compile-validators && tsc -p tests/tsconfig.json && mocha",
+    "test:node": "rimraf tests/compiled && pnpm compile-validators && tsc -p tests/tsconfig.json && c8 mocha",
```
jiyoontbd, line 86: what is `c8` for?

packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -1,94 +1,126 @@
...
+    const metadata: RfqMetadata = {
       ...opts.metadata,
-      kind       : 'rfq' as const,
+      kind       : 'rfq',
       id         : id,
       exchangeId : id,
       createdAt  : new Date().toISOString()
     }
 
-    // TODO: hash `data.payinMethod.paymentDetails` and set `private`
```
jiyoontbd, line 45: could we keep these `// TODO` if they aren't completed in this pr and are still to be done?

packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -1,94 +1,126 @@
...
+   * @throws if payinMethod in {@link Rfq.data} property `kind` cannot be validated against the provided offering's payinMethod kinds
+   * @throws if payinMethod in {@link Rfq.data} property `paymentDetails` cannot be validated against the provided offering's payinMethod requiredPaymentDetails
+   * @throws if payoutMethod in {@link Rfq.data} property `kind` cannot be validated against the provided offering's payoutMethod kinds
+   * @throws if payoutMethod in {@link Rfq.data} property `paymentDetails` cannot be validated against the provided offering's payoutMethod requiredPaymentDetails
    */
-  async verifyOfferingRequirements(offering: Offering | ResourceModel<'offering'>) {
-    if (offering.metadata.id !== this.offeringId) {
-      throw new Error(`offering id mismatch. (rfq) ${this.offeringId} !== ${offering.metadata.id} (offering)`)
+  async verifyOfferingRequirements(offering: Offering) {
+    if (offering.metadata.id !== this.data.offeringId) {
```
jiyoontbd, line 96: i don't understand how this worked previously without the `.data.`...

jiyoontbd: ooh i get it now coz we previously had [a bunch of getters](https://github.com/TBD54566975/tbdex-js/pull/156/files#diff-3c7fba4cf3225299da3d9b5bf8f36f2c3e0bfe87b86d14b4584f35b69c012906L162)

packages/protocol/src/parse.ts
```diff
@@ -0,0 +1,118 @@
...
+ * @param message - the message to parse. can either be an object or a string
+ * @returns {@link Message}
+ */
+export async function parseMessage(rawMessage: MessageModel | string): Promise<Message> {
+  const jsonMessage = rawToMessageModel(rawMessage)
+
+  let message: Message
+  switch(jsonMessage.metadata.kind) {
+    case 'rfq':
+      message = new Rfq(
```
jiyoontbd, line 24:
```suggestion
      message = await Rfq.parse(rawMessage)
```

just wondering if we can use `parse()` method you already wrote for each of the concrete message impls?

packages/protocol/tests/offering.spec.ts
```diff
@@ -1,64 +1,81 @@
...
-        Offering.validate(offering)
-        expect.fail()
-      } catch (e) {
-        expect(e.message).to.include('additional properties')
-      }
+      }).to.throw
     })
   })
 
+  // describe('validate', () => {
```
jiyoontbd, line 37: remove these commented out bits if not relevant anymore?

packages/protocol/src/parse.ts
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * @beta
+ *
+ * Parses the json message into a message instance.
+ * Performs format validation and an integrity check on the signature
+ * @param message - the message to parse. can either be an object or a string
+ * @returns {@link Message}
+ */
+export async function parseMessage(rawMessage: MessageModel | string): Promise<Message> {
```
jiyoontbd, line 18: thoughts on being able to call this like `MessageUtils.parseMessage()` as a static method? a good amount of our other API surfaces tend to follow this pattern i.e. `Rfq.parse()` or `Crypto.verify()`

jiyoontbd (approved):
soo awesome!! i _love_ the changes you've made to `protocol` (which in turn also affects `http-client` positively in simplifying types and tightening type constraints wherever possible). i also very much appreciate the small cleanups throughout this PR.  

i don't immediately love the api changes to `http-server` - i wish we could preserve some of our generic magic in `submit<T>` and `get<T>` so we don't have to write / update concrete impls if tbdex messages change in the future. but i think changes i have to make as a consumer won't be too bad.

packages/protocol/src/message-kinds/rfq.ts
jiyoontbd: looks like in other areas of this PR that we're doing away with the concept of `private` fields? not 100% sure tho

jiyoontbd:
oh, could you add a changeset? steps [here](https://github.com/TBD54566975/tbdex-js?tab=readme-ov-file#steps-for-a-new-release-publish)

packages/protocol/package.json
diehuxx: c8 produces a code coverage report when you run the tests

packages/http-client/src/client.ts
```diff
@@ -385,10 +383,10 @@ export type GetOfferingsOptions = {
   pfiDid: string
   filter?: {
     /** ISO 3166 currency code string */
-    payinCurrency?: OfferingData['payinCurrency']['currencyCode']
+    payinCurrency?: string
```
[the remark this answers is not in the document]
diehuxx: I'm mainly changing it to string because it's much more readable, similar to other type changes in this PR, I'm moving away from abstraction and "automated" typing toward more concrete and readable types.

packages/protocol/src/message-kinds/rfq.ts
diehuxx: My bad, I should have provided more context there. In [this proposal](https://github.com/TBD54566975/tbdex/pull/232) for how to handle private data, we're doing away with the notion of `private` as a top-level field in RFQs. This PR does not implement the proposal, but it does clean up `private` in preparation for implementing the proposal. I'll put the TODO back in.

packages/protocol/src/parse.ts
diehuxx: I'd prefer not to reuse `Rfq.parse` etc. `Rfq.parse()` does some of the same things that are already done in `parseMessage`, so we would be repeating operations (`.verify()`, and `rawToMessageModel`). While we could remove those operations from `parseMessage`, I think it would be less clear what `parseMessage` is doing by putting those things in a method. Also, i prefer to think of `Rfq.parse()` and `parseMessage` as part of the public API for this library, and that we use the constructors and instance methods when appropriate internally.

packages/protocol/src/parse.ts
diehuxx: Works for me. Instead of `MessageUtils`, I'm calling it `Parser` which matches the grammatical pattern of our `Validator` and `DidResolver`. `MessageUtils` seems a little too broad, plus this file contains utils for parsing resources in addition to messages.

packages/protocol/src/message-kinds/rfq.ts
diehuxx: Updated

diehuxx:
> i don't immediately love the api changes to http-server - i wish we could preserve some of our generic magic in submit<T> and get<T>

Fair enough. Here's my defense.

First, a petite polemic: Generics are a fatal temptation.  The DRYness that generics promise only breeds a thirst for even more generics. One drop may be invigorating, but we must not become addicted, lest we resent their absence.

In this PR I adopt the principle that anything that _can_ be accomplished clearly and readably without generics, _should_ be done without generics. The existing `get` and `submit` functions in `http-server` have largely the same appearance as the new `onSubmitRfq()`, `onGetOfferings`, etc.

For example,  `server.submit('rfq', rfqCallback)` looks pretty similar to `server.onSubmitRfq(rfqCallback)`. Both versions require the consumer to realize that they are specifying a callback for the `submit RFQ` endpoint. But the latter has a simpler implementation by avoiding  the use of generics. And, I'd argue that it's easier to document/understand that calling `server.onSubmitRfq(rfqCallback)` overwrites (rather than adds to) the original callback.

> so we don't have to write / update concrete impls if tbdex messages change

I'd argue this is an anti-pattern in the context of `tbdex/http-server`. There's a limited number of endpoints we're trying to support. If we add new tbdex messages and endpoints, we **should** have to add new methods to `TbdexHttpServer` and add unit tests accordingly. Being able to add new endpoints without needing to update `TbdexHttpServer` seems like a recipe for untested codepaths.

packages/http-server/src/http-server.ts
diehuxx: Yep that's right. See this comment for my reasoning https://github.com/TBD54566975/tbdex-js/pull/156#issuecomment-1930887063.

packages/protocol/tests/offering.spec.ts
diehuxx: upd8d

mistermoe:
fair @diehuxx i'll take maintainability and clarity over hipster api surface vibes. fwiw `server.submit('rfq', rfqCallback)` always felt kinda odd to me when i implemented it because i always read it as "i the server am submitting an rfq and then calling `rfqCallback`". whereas `server.onSubmitRfq` reads far more like "i the server, on receiving an rfq will call `rfqCallback`

looks like all that's left is addressing the tbdocs warnings i think?

mistermoe (approved):

mistermoe:
cc: @angiejones @chris-tbd

PR #163, opened 2024-02-08
Add external_id field

phoebe-lew:


KendallWeihe:
(I've been out of the loop so apologies if this is easily answered)

Should we add this to [the spec](https://github.com/TBD54566975/tbdex/tree/main/specs/protocol)?

Are there implications that the `external_id` will be echoed back to Alice in subsequent messages (such as Quote or OrderStatus)?

KendallWeihe:
lol literally the next [PR I pulled up to review was this](https://github.com/TBD54566975/tbdex/pull/235#event-11741294068), disregard the first question

phoebe-lew:
> Are there implications that the external_id will be echoed back to Alice in subsequent messages (such as Quote or OrderStatus)?

@KendallWeihe this is a good point actually...perhaps we should only echo back the IDs which were sent by alice?

phoebe-lew:
@KendallWeihe was thinking about this more...I think we can actually assume that the entire contents of a message (metadata + message body) are visible between Alice and the PFI. If the PFI doesn't want to expose an internal ID, it can simply not populate the value. It's likely that there are a bunch of IDs correlated to a single exchange (one for payout side, one for payin side at least), those IDs are in a separate context from tbdex messages. The PFI can map from tbdex message id to payment rail specific IDs in its own business logic.

KendallWeihe:
@phoebe-lew makes sense to me. I'm curious, was there a motivating use case for adding `external_id` to the protocol or are we thinking ahead to potential use cases?

jiyoontbd (approved):

packages/protocol/tests/generate-test-vectors.ts
```diff
@@ -31,7 +31,7 @@ const generateParseQuoteVector = async () => {
   const did = await DidKeyMethod.create()
   const quote = Quote.create({
     metadata: {
-      exchangeId : Message.generateId('rfq'),
+      exchangeId : Message.generateId('quote'),
```
diehuxx, line 34: `exchangeId` should actually be an RFQ id. The `id` of the Quote message gets set within `Quote.create()`

phoebe-lew:
> I'm curious, was there a motivating use case for adding external_id to the protocol or are we thinking ahead to potential use cases?

@KendallWeihe guess it's a potential use case, but letting clients pass in an arbitrary reference/id is pretty standard practice with most payment APIs.

KendallWeihe (approved):

PR #166, opened 2024-02-08
Make `pfiDid` required property in `TbdexHttpServer`

### LINKED ISSUE #161: Require `pfiDid` when `exchangesApi` option is passed to `TbdexHttpServer` constructor
When constructing the TbdexHttpServer a consumer should always pass a `pfiDid` anytime they pass an `exchangesApi` option. Right now the `pfiDid` option is marked as an optional field. 

The [default](https://github.com/TBD54566975/tbdex-js/blob/main/packages/http-server/src/http-server.ts#L45) is currently set to `did:ex:pfi`, and there are no guardrails to guarantee a consumer will pass their `pfiDid` as needed when instantiating the server with an `exchangesApi` but this will cause errors

kirahsapong:
closes #161

### DIFF
--- a/packages/http-server/src/http-server.ts
+++ b/packages/http-server/src/http-server.ts
@@ -39,11 +39,6 @@ type NewHttpServerOptions = {
   pfiDid: string
 }
 
-const defaults: Omit<NewHttpServerOptions, 'pfiDid'> = {
-  offeringsApi : fakeOfferingsApi,
-  exchangesApi : fakeExchangesApi
-}
-
 /**
  * TBDex HTTP Server powered by Express
  * @beta
@@ -76,12 +71,10 @@ export class TbdexHttpServer {
 
   constructor(opts?: NewHttpServerOptions) {
     this.callbacks = {}
-    opts = { ...defaults, ...opts }
-    const { offeringsApi, exchangesApi, pfiDid } = opts
 
-    this.exchangesApi = exchangesApi
-    this.offeringsApi = offeringsApi
-    this.pfiDid = pfiDid
+    this.exchangesApi = opts?.exchangesApi ?? fakeExchangesApi
+    this.offeringsApi = opts?.offeringsApi ?? fakeOfferingsApi
+    this.pfiDid = opts?.pfiDid ?? 'did:ex:pfi'
 
     // initialize api here so that consumers can attach custom endpoints
     const api = express()


packages/http-server/src/http-server.ts
```diff
@@ -36,13 +36,12 @@ type CallbackMap = {
 type NewHttpServerOptions = {
   offeringsApi?: OfferingsApi
   exchangesApi?: ExchangesApi,
-  pfiDid?: string
+  pfiDid: string
 }
 
-const defaults: NewHttpServerOptions = {
+const defaults: Omit<NewHttpServerOptions, 'pfiDid'> = {
```
diehuxx, line 42: If you turn on strict typescript (which I am currently working on doing for us), you'll notice that it doesn't like these lines in the constructor because -- for example -- `this.exchangesApi` cannot be `undefined`, but `NewHttpServerOptions.exchangesApi` _can_ be `undefined`.

```typescript
    opts = { ...defaults, ...opts }
    const { offeringsApi, exchangesApi, pfiDid } = opts
    this.exchangesApi = exchangesApi
    this.offeringsApi = offeringsApi
    this.pfiDid = pfiDid
```

We programmers can read the code and see that it should work fine. But I would argue it reveals a smell. Let's have the compiler work for us.

What do you think of removing this `defaults` object. and updating those lines in the constructor to look like this

```typescript
    this.exchangesApi = exchangesApi ?? fakeExchangesApi
    this.offeringsApi = offeringsApi ?? fakeExchangesApi
    this.pfiDid = pfiDid
```

.changeset/gentle-lies-provide.md
diehuxx, line 1: gentle lies really do provide

jiyoontbd (approved):
thank you for saving future us _hours_ of debugging

packages/http-server/src/http-server.ts
diehuxx: I don't see any difference. Did you mean to push a commit?

diehuxx: ah i see it there now. github says it got the commit a few minutes after i commented

mistermoe (approved):

diehuxx:
@kirahsapong To be clear, as of https://github.com/TBD54566975/tbdex-js/pull/166/commits/1d13ed22844927727733b918f9f549ca077d1e27, this PR does not change behavior at all?

diehuxx:
@kirahsapong Making sure I understand. As of https://github.com/TBD54566975/tbdex-js/pull/166/commits/1d13ed22844927727733b918f9f549ca077d1e27, this PR does not change any behavior or address #161?

diehuxx (approved):

PR #169, opened 2024-02-09
Stricten http-client

diehuxx:
Turn on strict typescript and do a few small cleanups. Also, remove `discoverPFIs` because it seems to be untested, undocumented, absent from the spec, and riddled with typescript looseness that it doesn't seem worth strictening.


phoebe-lew (approved):

jiyoontbd (approved):

michaelneale:
stricten is a great word so that is worth a 👍 alone.

KendallWeihe (approved):

PR #176, opened 2024-02-15
During request token verification, check that JWTHeader.kid includes the same DID as JWTClaims.iss

jiyoontbd:
wrote a check to resolve the issuer did and check against signing kid. also fixed some tbdocs "errors"


packages/http-client/tests/client.spec.ts
```diff
@@ -328,34 +329,36 @@ describe('client', () => {
     })
   })
 
-  describe('verifyRequestToken', () => {
+  describe.only('verifyRequestToken', () => {
```
jiyoontbd, line 332: reminder to me to remove `.only`

packages/http-client/tests/client.spec.ts
```diff
@@ -387,16 +390,37 @@ describe('client', () => {
...
-        const requestToken = await createRequestTokenFromPayload(payload)
+        const requestToken = await createRequestToken(payload)
         await TbdexHttpClient.verifyRequestToken({ requestToken, pfiDid: pfiPortableDid.did })
         expect.fail()
       } catch(e) {
         expect(e).to.be.instanceof(RequestTokenAudienceMismatchError)
         expect(e.message).to.include('Request token contains invalid audience')
       }
     })
+    it.only('throws RequestTokenIssuerSignerMismatchError if request token header kid does not match up with request token claimset iss', async () => {
```
jiyoontbd, line 401: reminder to me to remove `.only`

packages/http-client/src/client.ts
```diff
@@ -306,8 +307,16 @@ export class TbdexHttpClient {
       throw new RequestTokenAudienceMismatchError({ message: 'Request token contains invalid audience. Expected aud property to be PFI DID.' })
     }
 
-    // TODO: check iss against signer DID
-    return requestTokenPayload.iss!
+    const signerKid = requestTokenHeader.kid!
+    const issuer = requestTokenPayload.iss!
+    const didDocument = await resolveDid(issuer)
```
KendallWeihe, line 312: Why also resolve the DID Doc? As opposed to just `if (!signerKid.includes(issuer))`

KendallWeihe (approved):

KendallWeihe:
Oh @jiyoontbd I thought more about this concept (that the `kid` ought to equal the `iss`). I think this is fine (and maybe always will be?) but I think there is a set of circumstances where this can't be the case wherein the private key owner _delegates_ issuance to another entity -- in which case, the private key owner would be the signer (and hence have the value of the `kid`) but the issuer would be some other entity, of whom the private key owner has granted access to issue on their behalf.

jiyoontbd:
> ...I think there is a set of circumstances where this can't be the case wherein the private key owner _delegates_ issuance to another entity -- in which case, the private key owner would be the signer (and hence have the value of the `kid`) but the issuer would be some other entity, of whom the private key owner has granted access to issue on their behalf.

cc @decentralgabe is this how delegate keys would work in terms of creating a JWT (guess it applies here as much as it applies to VCs)?

i say right now we can make an assumption (can add comment in this part of the code) that we expect the signer and the issuer to be the same entity/DID.

KendallWeihe:
> i say right now we can make an assumption (can add comment in this part of the code) that we expect the signer and the issuer to be the same entity/DID.

I think this is my take too... because I'd imagine the tbdex spec will me ruleset prescriptions on this matter which more-oft-than-not how you have it now it would be the case

KendallWeihe:
> the tbdex spec

this actually raises a point... should we open a ticket on [the tbdex backlog](https://github.com/TBD54566975/tbdex/issues) to define this rule?

decentralgabe:
@jiyoontbd @KendallWeihe just commenting from the perspective of a DID Document, there is a _controller_ property to signify a key is controlled by another entity (whether it be another DID, or another URI of some sort).

My recommendation is that `kid` does not equal `iss`, where `iss` is `did:example:abcd` and `kid` is `did:example:abcd#key-1` this is referred to as a fully qualified key ID.

diehuxx (approved):
Looks good! Thanks for cleaning up the TBDocs errors. Feels good to see zero issues. Regarding, the assumption about whether `kid` always includes `iss`, I'd say it's best to go with that assumption for this PR and address again later after we revisit the spec rules.

diehuxx:
> this actually raises a point... should we open a ticket on [the tbdex backlog](https://github.com/TBD54566975/tbdex/issues) to define this rule?

https://github.com/TBD54566975/tbdex/issues/238

KendallWeihe:
@decentralgabe regarding the distinction between `kid` and `iss`, @jiyoontbd solved that with an `includes()` conditional rather than an equals conditional 

```typescript
if (!signerKid.includes(issuerDid)) { ... }
```

@diehuxx thanks for opening the ticket! Cross sharing the comment I made over there, basically IMO this is fine but at some point we may consider adding support for delegation

> _I think this comes to whether or not we should support delegation. The PR that is linked in the description enforces a verification rule which does not allow delegation, which I think is fine for the time being, but at some point we may want to extend support for delegation._

KendallWeihe (approved):
stamping again to let you know still lgtm!

jiyoontbd:
blocker: the commit hash once this PR is merged in tbdex should be the commit hash pointer in this PR: https://github.com/TBD54566975/tbdex/pull/247

jiyoontbd:
jiyoon todo: open a PR against tbdex-kt repo and pull in the new commit from https://github.com/TBD54566975/tbdex/pull/247 when merged

KendallWeihe:
> blocker: the commit hash once this PR is merged in tbdex should be the commit hash pointer in this PR: https://github.com/TBD54566975/tbdex/pull/247

@jiyoontbd are you waiting on the test vectors from that PR? (I ask because it appears test vectors are passing so I know I'm missing something)

PR #181, opened 2024-02-21
Clean up unused and extraneous code

diehuxx:
* Remove `DevTools.createDid()` and replace with `DidDht.create()` or `DidJwt.create()`. Based on rule of thumb that we should only use DevTools if the alternative is significantly more verbose.
* Remove dead code and update comments in `packages/protocol/src/did-resolver.ts` and `packages/protocol/src/crypto.ts`


KendallWeihe:
Changeset notes?

KendallWeihe (approved):
this is excellent! 

(I kinda think we may be able to axe the dev tools altogether, but maintain the dummy data for test purposes)

mistermoe (approved):
Love this

diehuxx:
> (I kinda think we may be able to axe the dev tools altogether, but maintain the dummy data for test purposes)

I'd be down. The only reason we're exporting devtools from `tbdex/protocol` is we're using those same fixtures in the http packages. The easiest alternative is to copypaste the test fixtures into all three packages and don't export the fixtures from any of them. I can cover that in a follow up PR.

PR #190, opened 2024-02-29
Use msw to test http-client requests

diehuxx:
TL;DR I'm using `msw` to write tests, but I'm using `msw` weirdly, but I have good reasons.

## Draft Phrase Preamble
Before I dive in and write ~20 tests using this approach, I want to get some soft consensus that people like testing the http-client this way. I tried several different libraries for mocking `fetch` and testing `fetch` requests, and honestly don't like any of them, but I've settled on an approach that works for our use case.

## Overview
Our current tests in the http-client are brittle. We test that we can call methods like `TbdexHttpClient.sendMessage` to make sure that they do not `throw`, but we have no assertions that we are actually sending a message to the correct url with the correct request structure. The aim of this PR is to add those tests.

We select `msw` to accomplish this because `msw` has allows us to record requests to an endpoint and `msw` supports a wider variety of environments that other libraries.

## Testing libraries comparison
I'll start by saying I settled on `msw`, but deliberately going against some of its recommended best practices. More detail below.

### nock
I found nock the most ergonomic for setting up one-time request interceptors. Its API was exactly what I wanted. Sadly it has limited environment support, and I don't want to have to re-write all these tests again in a month when nock inevitably doesn't cut it.

### Polly.js
I had high hopes when I saw Netflix built it, but the docs are all over the place and it's pretty complex to set up. The [quickstart](https://netflix.github.io/pollyjs/#/quick-start) is broken and uses deprecated APIs, and finally saw that [the adapter I needed was built on top of `nock` anyway](https://netflix.github.io/pollyjs/#/adapters/node-http). 

### msw
**Rant**: I tried `msw` first, hated it, then came crawling back. I don't like that `msw` feels like reimplementing the server in the test files. I REALLY don't like having a separate `handlers.js` where all the default ~mocks~ request handlers go. [Kent C Dodds](https://kentcdodds.com/blog/stop-mocking-fetch) seems to love how short it makes his tests, but personally I find it hard to read what the test is actually testing for. 

To better accommodate ~my neuroses~ our use case, these tests deliberately go against some of MSW's officially recommended best practices. In particular, these tests make [request assertions](https://mswjs.io/docs/best-practices/avoid-request-assertions) . MSW's reasoning is that request assertions tend to test implementation rather than behavior. This makes sense for apps where one entity controls (and can freely make changes to) both the client and server. However, we are implementing a spec, and must be sure that our client adheres precisely to the spec in the requests' url, header, and body.

Though other libraries (e.g. nock) supply assertions for request url, header, and body matching out of the box, we choose MSW because it has support for a wide variety of environments. To adapt it for our use case, we make extensive use of `server.use(http.get({ once: true, ... }))` as a form of assertion.

### DIFF
--- a/packages/http-client/tests/client.spec.ts
+++ b/packages/http-client/tests/client.spec.ts
@@ -1,421 +1,421 @@
-import { expect } from 'chai'
-import { DidDht, DidJwk, BearerDid } from '@web5/dids'
-import { TbdexHttpClient, requestTokenRequiredClaims } from '../src/client.js'
-import {
-  RequestError,ResponseError,
-  InvalidDidError,
-  MissingServiceEndpointError,
-  RequestTokenMissingClaimsError,
-  RequestTokenAudienceMismatchError,
-  RequestTokenVerificationError,
-  RequestTokenSigningError,
-  RequestTokenIssuerSignerMismatchError
-} from '../src/errors/index.js'
-import { DevTools } from '@tbdex/protocol'
-import * as sinon from 'sinon'
-import { JwtHeaderParams, JwtPayload } from '@web5/crypto'
-import { Convert } from '@web5/common'
-import { Jwt } from '@web5/credentials'
+// import { expect } from 'chai'
+// import { DidDht, DidJwk, BearerDid } from '@web5/dids'
+// import { TbdexHttpClient, requestTokenRequiredClaims } from '../src/client.js'
+// import {
+//   RequestError,ResponseError,
+//   InvalidDidError,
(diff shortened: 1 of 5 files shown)


packages/http-client/tests/client.spec.ts
diehuxx, line 1: Temporarily commenting this file out because it includes stubs that bleed over into other tests.

diehuxx:
Closing since I don't have time to get this ready for review, and we have an issue to track it. https://github.com/TBD54566975/tbdex-js/issues/197

PR #191, opened 2024-03-01
Add protocol version

### LINKED ISSUE #172: tbdex version changes
original spec issue: https://github.com/TBD54566975/tbdex/issues/159

https://hackmd.io/TucOUP8UQt-XKj8G9UsJ2Q?both

- [x] add version to message metadata
- [x] add version of resource metadata
- [ ] add supported versions/list PFI details endpoint

kirahsapong:
closes #172

### DIFF
--- a/packages/protocol/tests/exchange.spec.ts
+++ b/packages/protocol/tests/exchange.spec.ts
@@ -18,9 +18,8 @@ describe('Exchange', () => {
 
     rfq = Rfq.create({
       metadata: {
-        from     : aliceDid.uri,
-        to       : pfiDid.uri,
-        protocol : '1.0'
+        from : aliceDid.uri,
+        to   : pfiDid.uri
       },
       data: await DevTools.createRfqData()
     })
@@ -118,8 +117,7 @@ describe('Exchange', () => {
         metadata: {
           from       : pfiDid.uri,
           to         : aliceDid.uri,
-          exchangeId : Message.generateId('rfq'),
-
+          exchangeId : Message.generateId('rfq')
         },
         data: DevTools.createQuoteData()
       })
@@ -341,9 +339,8 @@ describe('Exchange', () => {
 
       const rfq = Rfq.create({
         metadata: {
-          from     : aliceDid.uri,
-          to       : pfiDid.uri,
-          protocol : '1.0'
+          from : aliceDid.uri,
+          to   : pfiDid.uri
         },
         data: await DevTools.createRfqData()
       })
--- a/packages/protocol/tests/rfq.spec.ts
+++ b/packages/protocol/tests/rfq.spec.ts
@@ -176,7 +176,7 @@ describe('Rfq', () => {
       offeringData.requiredClaims = undefined
       const offering = Offering.create({
         metadata: {
-          from: pfi.uri,
+          from: pfi.uri
         },
         data: offeringData
       })
@@ -453,7 +453,7 @@ describe('Rfq', () => {
       const pfi = await DidJwk.create()
 
       const offering = Offering.create({
-        metadata : { from: pfi.uri,  },
+        metadata : { from: pfi.uri },
         data     : offeringData,
       })
       await offering.sign(pfi)
@@ -471,7 +471,7 @@ describe('Rfq', () => {
       const rfq = Rfq.create({
         metadata: {
           from : alice.uri,
-          to   : pfi.uri
+          to   : pfi.uri,
         },
         data: rfqData,
       })
@@ -500,7 +500,7 @@ describe('Rfq', () => {
       rfqData.claims = [vcJwt]
 
       const rfq = Rfq.create({
-        metadata : { from: aliceDid.uri, to: 'did:ex:pfi',  },
+        metadata : { from: aliceDid.uri, to: 'did:ex:pfi' },
         data     : rfqData
       })
 


packages/protocol/src/message.ts
```diff
@@ -44,6 +46,11 @@ export abstract class Message {
     return typeid(messageKind).toString()
   }
 
+  /** Gets the 'x.x' major/minor formatted version based on the current protocol release version */
+  static getProtocolVersion(): `${number}` {
```
diehuxx, line 50: Parsing this as a number is going to lead to weird comparisons, e.g. Version 1.2 > 1.12. What do you think of returning a string?

packages/http-server/src/request-handlers/create-exchange.ts
```diff
@@ -46,6 +46,14 @@ export async function createExchange(req: Request, res: Response, options: Creat
     return
   }
 
+  if (offering.metadata.protocol !== rfq.metadata.protocol) {
```
diehuxx, line 49: What do you think of moving this check into `Rfq.verifyOfferingRequirements()`?

packages/protocol/src/types.ts
```diff
@@ -130,6 +132,8 @@ export type MessageMetadata = {
   createdAt: string
   /** Arbitrary ID for the caller to associate with the message. Optional */
   externalId?: string
+  /** Version of the protocol in use (x.x format). Must be consistent with all other messages in a given exchange */
+  protocol: `${number}`
```
diehuxx, line 136: Ditto: https://github.com/TBD54566975/tbdex-js/pull/191/files#r1510162565

What do you think of having `protocol` be a string?

packages/protocol/src/message.ts
diehuxx: Oh wait this is the version of the SDK, not the TBDex version. I think those two versions will be different, no? What are we using the SDK version for within the SDK?

diehuxx: TIL template literals. thanks!

packages/protocol/src/message-kinds/rfq.ts
diehuxx, line 1: Suggestion: populate `protocol` within `.create()` rather than passing it as an parameter. This package only supports one possible version currently, so it could simplify DX.

.changeset/stale-hats-run.md
jiyoontbd, line 1: 🎩 🏃

packages/protocol/src/message-kinds/rfq.ts
mistermoe: agreed. i think `protocol` should be set internally and optionally overridable

packages/protocol/src/message.ts
KendallWeihe: wow TIL template literal as well, very cool

packages/protocol/src/message-kinds/rfq.ts
phoebe-lew: So, default to 1.0 for now with no validation? Or hardcode the versions somewhere the way Kirah previously had it?

diehuxx: I'm not sure what you mean by no validation, but either option sounds sufficient to me

mistermoe: default to `1.0` with no validation feels like easiest path forward right now. wdyt @phoebe-lew ?

phoebe-lew: @mistermoe agree that's easiest.

@diehuxx if overridable with no validation, someone could put in `666.6` or `not even a number` as the version

mistermoe: we can create an issue to track validating `protocol` and tackle post 1.0. i think we'll have a better sense for what "validation" entails beyond just the actual value (e.g. is this message conformant to the version specified) when we're closer to introducing a change to the protocol

diehuxx: @mistermoe ah i see, thanks! agreed let's not do validation for now and we'll follow up on it later

packages/protocol/src/message-kinds/close.ts
```diff
@@ -8,7 +8,7 @@ import { Parser } from '../parser.js'
  */
 export type CreateCloseOptions = {
   data: CloseData
-  metadata: Omit<CloseMetadata, 'id' | 'kind' | 'createdAt'>
+  metadata: Omit<CloseMetadata, 'id' | 'kind' | 'createdAt' | 'protocol'> & { protocol?: CloseMetadata['protocol'] }
```
diehuxx, line 11: Non-blocking: I kinda hate indexed access type syntax in typescript. Instead I prefer to just use the underlying type, in this case,
```
`${number}`
```
```suggestion
  metadata: Omit<CloseMetadata, 'id' | 'kind' | 'createdAt' | 'protocol'> & { protocol?: `${number}` }
```

Up to you if you feel like doing this. Just a personal preference on my part.

diehuxx (approved):

packages/protocol/src/message-kinds/close.ts
phoebe-lew: Curious what you don't like about it! the way it looks, or is there something about the way it works that you dislike?

diehuxx: it's a readability thing. it's an abstraction that i personally don't find helpful. i don't like needing to have to look at a separate file or type to figure out what i'm actually working with, especially if it's as simple as a number or string.

phoebe-lew: Oh got it. I'm leaning towards leaving it as is because theoretically, the types could drift between `CloseMetadata` and this field if it's not referencing `CloseMetadata`.

jiyoontbd (approved):

packages/protocol/src/message-kinds/rfq.ts
jiyoontbd: issue created: https://github.com/TBD54566975/tbdex/issues/270

Issue #187, opened 2024-02-26
Split `TbdexHttpClient.sendMessage()` into separate methods `createExchange()`, `submitOrder()`, and `submitClose()`


diehuxx:
`sendMessage` is misleading for two reasons.
1. The client should not be able to send just any message to a PFI. They should only be able to send RFQs, Orders, and Closes. Other message kinds are created and sent by the PFI.
2. `sendMessage` sends to several different endpoints with different options. We are using clever typescript to enforce that the correct options are available for each message type, but these clever tricks result in less obvious usage for consumers of `http-client`. This is part of my [crusage against generics](https://github.com/TBD54566975/tbdex-js/pull/156#issuecomment-1930887063).

Current `sendMessage` method signature:
```typescript
export type SendMessageOptions<T extends Message> = {
  message: T
  replyTo?: T extends Rfq ? string : never
}

static async sendMessage<T extends Message>(opts: SendMessageOptions<T>): Promise<void>
```

Proposed method signatures:
```typescript
static async createExchange(rfq: Rfq, opts?: { replyTo?: string }): Promise<void>

static async submitOrder(order: Order): Promise<void>

static async submitClose(close: Close): Promise<void>
```

The proposed methods are less DRY because DRY APIs are not very tasty.

diehuxx:
@angiejones This would constitute a breaking change in the `http-client` implementation. It would make it harder to use the client incorrectly.

angiejones:
This should be opened against the protocol, yes? Since it would need to be implemented in all SDKs. Would we have time to do this before v1 (end of the month)?

diehuxx:
@angiejones Good call, I opened an issue in the `tbdex` repo to track across implementations https://github.com/TBD54566975/tbdex/issues/254. This issue doesn't affect the TBDex spec, but since it affects multiple of our implementations, makes sense to track it that way.

Regarding timeline, yes this will be done by v1. The javascript and kotlin implementations are ready for review. I will keep an eye on the PRs to close them out as soon as possible.

diehuxx:
@angiejones KT and JS are both merged

PR #182, opened 2024-02-21
Remove exchangeId from submit endpoint paths

diehuxx:
Implements spec change https://github.com/TBD54566975/tbdex/pull/243 https://github.com/TBD54566975/tbdex/issues/225

PR #192, opened 2024-03-04
Split client sendMessage into separate methods

diehuxx:
Implements: https://github.com/TBD54566975/tbdex/issues/254 which is the same as https://github.com/TBD54566975/tbdex-js/issues/187

### DIFF
--- a/packages/http-client/src/client.ts
+++ b/packages/http-client/src/client.ts
@@ -2,7 +2,9 @@ import type { JwtPayload } from '@web5/crypto'
 import type { ErrorDetail } from './types.js'
 import type { DidDocument, BearerDid } from '@web5/dids'
 import {
+  Close,
   MessageModel,
+  Order,
   Parser,
   Rfq,
 } from '@tbdex/protocol'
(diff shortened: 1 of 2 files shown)


packages/http-client/tests/client.spec.ts
```diff
@@ -108,7 +110,176 @@ describe('client', () => {
...
+      fetchStub.rejects({message: 'Failed to fetch on URL'})
+
+      const close = Close.create({
+        metadata: {
+          from       : aliceDid.uri,
+          to         : pfiDid.uri,
+          exchangeId : Message.generateId('rfq')
+        },
+        data: {
+          reason: 'Closed for the day. Gone fishin'
```
jiyoontbd, line 212: 🎣

packages/http-client/src/client.ts
```diff
@@ -55,30 +57,63 @@ export const requestTokenRequiredClaims = ['aud', 'iss', 'exp', 'iat', 'jti']
...
+   * @throws if recipient DID resolution fails
+   * @throws if recipient DID does not have a PFI service entry
+   */
+  static async createExchange(rfq: Rfq, opts?: { replyTo?: string }): Promise<void> {
+    await rfq.verify()
+
+    const { to: pfiDid, exchangeId } = rfq.metadata
+    const requestBody = JSON.stringify({ rfq, replyTo: opts?.replyTo })
+
+    await TbdexHttpClient._sendMessage(pfiDid, `/exchanges/${exchangeId}/rfq`, requestBody)
```
jiyoontbd, line 75: i guess this will need to be updated once our debate in #182 is resolved?

jiyoontbd (approved):

packages/http-client/src/client.ts
diehuxx: Yep, I'll update it in another PR once we make a decision.

packages/http-client/src/client.ts
```diff
@@ -55,30 +57,63 @@ export const requestTokenRequiredClaims = ['aud', 'iss', 'exp', 'iat', 'jti']
...
+    const { to: pfiDid, exchangeId } = close.metadata
 
-    await message.verify()
+    const requestBody = JSON.stringify(close)
 
-    const { to: pfiDid, exchangeId, kind } = message.metadata
+    await TbdexHttpClient._sendMessage(pfiDid, `/exchanges/${exchangeId}/close`, requestBody)
+  }
+
+  private static async _sendMessage(pfiDid: string, path: string, requestBody: string): Promise<void> {
```
KendallWeihe, line 111: Just an idea, if we actually want to make this private at runtime (as opposed to build time) then 

```suggestion
  static async #sendMessage(pfiDid: string, path: string, requestBody: string): Promise<void> {
```

Not sure it matters much either way

packages/http-client/src/client.ts
```diff
@@ -55,30 +57,63 @@ export const requestTokenRequiredClaims = ['aud', 'iss', 'exp', 'iat', 'jti']
...
     const pfiServiceEndpoint = await TbdexHttpClient.getPfiServiceEndpoint(pfiDid)
-    const apiRoute = `${pfiServiceEndpoint}/exchanges/${exchangeId}/${kind}`
+    const apiRoute = `${pfiServiceEndpoint}${path}`
 
     let response: Response
     try {
-      let requestBody
-      if (message.metadata.kind == 'rfq') {
```
KendallWeihe, line 77: yay for less conditional logic embedded within a function for one-off use cases!

KendallWeihe (approved):
woot woot!

packages/http-client/src/client.ts
diehuxx: Since we're using `private` rather than `#` syntax throughout the SDK, I'd prefer to leave it for now. But if we want to update it throughout the SDK, I'm open to doing that in another PR

diehuxx: I decided to remove the underscore from the method name though.

PR #203, opened 2024-03-18
build and publish snapshot releases on every merge to main

finn-tbd:
This is part of a project being undertaken by OSP to help new code make it's way to TBD's downstream projects faster. It creates a snapshot release for each merge to main, which is published to our Artifactory* registry. Once this is done, downstream projects (developer.tbd.website) can pull in new tbdex-js features before a real release has been produced by depending on the snapshot version.

*Artifactory: behaves like npmjs.com, but isn't real npmjs.com so the high number of releases this creates will not spam up our npmjs.com page.

### DIFF
--- a/.github/workflows/release-snapshot.yml
+++ b/.github/workflows/release-snapshot.yml
@@ -1,0 +1,87 @@
+name: release snapshot to jfrog
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+# Allow only one concurrent deployment, but do NOT cancel in-progress runs as
+# we want to allow these release deployments to complete.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  publish-npm:
+    name: Snapshot Publish
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
+        with:
+          fetch-depth: 0
+          submodules: "true"
+
+      # https://cashapp.github.io/hermit/usage/ci/
+      - name: Init Hermit
+        uses: cashapp/activate-hermit@31ce88b17a84941bb1b782f1b7b317856addf286 #v1.1.0
+        with:
+          cache: "true"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release-snapshot.yml
```diff
@@ -0,0 +1,87 @@
...
+          REGISTRY: https://blockxyz.jfrog.io/artifactory/api/npm/tbd-oss-snapshots-npm/
+        run: |
+          set -exuo pipefail
+          cd packages/protocol
+
+          base_version=$(jq -r .version package.json)
+          commits_since_version_bump=$(git rev-list HEAD $(git describe --tags --abbrev=0 HEAD) --count)
+          latest_commit="$(git rev-parse HEAD)"
+
+          snapshot_version="${base_version}-SNAPSHOT.${commits_since_version_bump}-${latest_commit:0:7}"
```
KendallWeihe, line 55: instead of `latest_commit:0:8` you could also use the `--short` flag in the `git` command

```shell
git rev-parse --short HEAD
```

.github/workflows/release-snapshot.yml
```diff
@@ -0,0 +1,87 @@
...
+
+          # check if that snapshot version has already been published
+          if npm view --registry "${REGISTRY}" "@tbdex/protocol@${snapshot_version}" > /dev/null; then
+              echo "release for @tbdex/protocol@${snapshot_version} already exists, not re-publishing"
+              exit 0
```
KendallWeihe, line 60: probably this would be considered an error case, right? so then `exit 1`

finn-tbd: that's true, actually come to think of it this check doesn't need to exist for tbdex-js. It's left over from web5-js where the subpackages are versioned separately

.github/workflows/release-snapshot.yml
```diff
@@ -0,0 +1,87 @@
...
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+# Allow only one concurrent deployment, but do NOT cancel in-progress runs as
+# we want to allow these release deployments to complete.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
```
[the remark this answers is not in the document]
finn-tbd: good point, left over from web5-js

finn-tbd:
Thanks for catching those. I think all of them were a result of me copying this over from web5-js and modifying it minimally to fit the different release expectations. I've addressed the three comments.

jiyoontbd (approved):

Issue #196, opened 2024-03-13
Restructure http endpoints according to latest design


diehuxx:
Design specified in this [comment here](https://github.com/TBD54566975/tbdex/issues/225#issuecomment-1986057713) in an issue in the spec repo

PR #204, opened 2024-03-18
#196 Restructure http endpoints according to latest design

diehuxx:
Addresses https://github.com/TBD54566975/tbdex-js/issues/196
New design here https://github.com/TBD54566975/tbdex/issues/225#issuecomment-1986057713

### DIFF
--- a/packages/http-server/src/http-server.ts
+++ b/packages/http-server/src/http-server.ts
@@ -1,11 +1,12 @@
 import type {
   OfferingsApi,
   ExchangesApi,
-  SubmitRfqCallback,
+  SubmitRfqCallback as CreateExchangeCallback,
   SubmitOrderCallback,
   SubmitCloseCallback,
   GetExchangesCallback,
   GetOfferingsCallback,
+  GetExchangeCallback,
 } from './types.js'
 
 import type { Express, Request, Response } from 'express'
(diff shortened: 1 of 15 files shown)


packages/http-server/src/http-server.ts
```diff
@@ -140,39 +153,41 @@ export class TbdexHttpServer {
...
         offeringsApi,
         exchangesApi,
       })
     )
 
-    this.api.post('/exchanges/:exchangeId/order', (req: Request, res: Response) =>
-      submitOrder(req, res, {
-        callback: this.callbacks['order'],
-        exchangesApi
+    this.api.post('/exchanges/:exchangeId', (req: Request, res: Response) =>
```
KendallWeihe, line 164: This is supposed to be a `put` right?

packages/http-server/tests/get-exchange.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    expect(respBody['errors']).to.exist
+    expect(respBody['errors'].length).to.equal(1)
+    expect(respBody['errors'][0]['detail']).to.include('Malformed Authorization header')
+  })
+
+  it('returns 404 if no matching exchange is found', async () => {
+    const rfq = await DevTools.createRfq({ sender: alice, receiver: pfi })
+
+    // Deliberately omit rfq from ExchangesApi
+    // (api.exchangesApi as InMemoryExchangesApi).addMessage(rfq)
```
KendallWeihe, line 70: needed?

packages/http-server/tests/submit-message.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    const pfiDid = await DidJwk.create()
+
+    const order = Order.create({
+      metadata: {
+        from       : aliceDid.uri,
+        to         : pfiDid.uri,
+        exchangeId : Message.generateId('rfq'),
+      },
+    })
+    // deliberately omit await order.sign(aliceDid)
```
KendallWeihe, line 50: needed?

KendallWeihe:
Should we add changesets?

packages/http-server/tests/get-exchange.spec.ts
diehuxx: Not needed. I do this if I'm deliberately omitting an action from the test. If it hinders readability rather than helps I can remove.

KendallWeihe (approved):
Would like a response to my comment regarding changesets, but other than that this LGTM

diehuxx:
> Should we add changesets?

done

packages/http-server/src/http-server.ts
```diff
@@ -1,33 +1,37 @@
 import type {
   OfferingsApi,
   ExchangesApi,
-  SubmitRfqCallback,
+  SubmitRfqCallback as CreateExchangeCallback,
```
jiyoontbd, line 4: is the type actually called this or did vscode do this automatically for you and you meant to actually import `CreateExchangeCallback`?

phoebe-lew (approved):

packages/http-server/src/http-server.ts
diehuxx: Good catch, that was VSCode. I'll update

PR #212, opened 2024-03-25
Add balance resource and todos

### LINKED ISSUE #138: Implement GET account balance endpoint and client method
per spec: https://github.com/TBD54566975/tbdex/pull/221

kirahsapong:
closes #138 

- adds `Balance` resource and protocol tests
- adds client `getBalance` method and tests
- adds server `GET /balance` endpoint, handlers and tests

### DIFF
--- a/packages/protocol/build/compile-validators.js
+++ b/packages/protocol/build/compile-validators.js
@@ -69,4 +69,4 @@ const moduleCode = replaceRequireWithImport(generatedCode)
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
 
 await mkdirp(path.join(__dirname, '../generated'))
-fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.cjs'), moduleCode)
+fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.js'), moduleCode)
--- a/packages/protocol/src/validator.ts
+++ b/packages/protocol/src/validator.ts
@@ -1,6 +1,6 @@
 import type { ErrorObject } from 'ajv'
 // validator functions are compiled at build time. check ./build/compile-validators.js for more details
-import * as compiledValidators from '../generated/compiled-validators.cjs'
+import * as compiledValidators from '../generated/compiled-validators.js'
 
 /**
  * validates the payload against a json schema identified by name


packages/http-server/src/http-server.ts
```diff
@@ -29,18 +33,27 @@ type CallbackMap = {
...
+ * @param opts.balancesApi Optionally accepts a {@link BalancesApi}. Defaults to an {@link InMemoryBalancesApi} which supports additional methods.
+ * @param opts.pfiDid Required if instantiating the HttpServer with options. Else, defaults to an arbitrary string for example purposes only.
  * @beta
  */
 type NewHttpServerOptions = {
   offeringsApi?: OfferingsApi
   exchangesApi?: ExchangesApi,
+  balances?: {
+    balancesApi?: BalancesApi,
+  }
```
KendallWeihe, line 56: why not? (like the others)

```suggestion
  balancesApi?: BalancesApi,
```

KendallWeihe: is this relevant to the line of code in the constructor?

```ts
this.balancesApi = opts?.balances ? opts?.balances.balancesApi ?? new InMemoryBalancesApi() : undefined
```

so the idea being PFIs may choose to not support the balances endpoint at all yeah? in that case you could make the default (of my proposal) `undefined` instead of `InMemoryBalancesApi`

packages/http-server/src/request-handlers/get-balances.ts
```diff
@@ -0,0 +1,21 @@
+import { Request, Response } from 'express'
+import type { GetBalancesCallback, BalancesApi } from '../types.js'
+
+type GetBalancesOpts = {
+  callback?: GetBalancesCallback
+  balancesApi: BalancesApi
+}
+
+export async function getBalances(request: Request, response: Response, opts: GetBalancesOpts): Promise<void> {
+  const { callback, balancesApi } = opts
```
KendallWeihe, line 10: I believe balance endpoint needs to be authz locked down, right? good inspiration from the `request-handlers/get-exchanges.ts`

packages/http-server/src/http-server.ts
KendallWeihe: gotcha so if the developer wants to implement the balances endpoint and wants to use the `InMemoryBalancesApi` then they would write

```ts
new TbdexHttpServer({ balances: { }, pfiDid: 'did:ex:pfi' })
```

---

but in this case, the developer knows they want to implement it so is that better than this? (assume the default of `NewHttpServerOptions.balancesApi` is `undefined` instead of the in memory instance)

```ts
new TbdexHttpServer({ balancesApi: InMemoryBalancesApi(), pfiDid: 'did:ex:pfi' })
```

KendallWeihe (approved):
two open comments but neither are blockers

awesome work here!

packages/protocol/build/compile-validators.js
```diff
@@ -51,20 +53,20 @@ const generatedCode = standaloneCode(validator)
...
 }
 const moduleCode = replaceRequireWithImport(generatedCode)
 
 
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
 
 await mkdirp(path.join(__dirname, '../generated'))
-fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.js'), moduleCode)
\ No newline at end of file
+fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.cjs'), moduleCode)
```
jiyoontbd, line 72: could you enlighten me on what this change from `.js` to `.cjs` does here?

packages/protocol/tests/balance.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+
+      try {
+        await balance.verify()
+        expect.fail()
+      } catch(e) {
+        expect(e.message).to.include(`must have required property 'signature'`)
+      }
+    })
+
+    xit('throws an error if signature is not a valid compact JWS')
```
jiyoontbd, line 97: will these become real tests in this PR or future PR?

tbdex
jiyoontbd, line 1: is there a PR for this change in the submodule?

jiyoontbd (approved):

tbdex
jiyoontbd: oh maybe you just have pulled in the latest commit

packages/protocol/build/compile-validators.js
diehuxx: I'm guessing `git merge main` will remove this change. We were dealing with some `cjs` related stuff in [this PR](https://github.com/TBD54566975/tbdex-js/pull/213#issuecomment-2021767436), but we were able to keep it as `.js` in the end.

packages/http-server/src/http-server.ts
```diff
@@ -79,6 +93,7 @@ export class TbdexHttpServer {
 
     this.exchangesApi = opts?.exchangesApi ?? new InMemoryExchangesApi()
     this.offeringsApi = opts?.offeringsApi ?? new InMemoryOfferingsApi()
+    this.balancesApi = opts?.balancesApi ?? undefined
```
diehuxx, line 96:
```suggestion
    this.balancesApi = opts?.balancesApi ?? new InMemoryBalancesApi()
```

packages/protocol/tests/balance.spec.ts
diehuxx: We have these as `xit`s for Offering and Rfq also. They're a remnant from before we had almost any tests in this repo, and honestly I'm not sure if these tests really belong inside message-kind-specific unit tests. They're really testing that `Crypto.verify()` works as expected. For the sake of this PR, I'm comfortable with leaving this as `xit`s or even omitting them.

diehuxx (approved):

packages/http-server/src/http-server.ts
KendallWeihe: the balances endpoint is not-implemented by default (this?.privateData?.salt === undefined)

> This endpoint is OPTIONAL. It is only relevant for PFIs which expose a stored balance functionality either for institutional or retail customers.

PR #213, opened 2024-03-26
Recent small spec changes

diehuxx:
Implement spec changes from this PR: https://github.com/TBD54566975/tbdex/pull/269

There are currently no up-to-date test vectors for RFQ and Offerings. We can't create functioning test vectors until 2 tbdex SDKs have implemented these changes. I would prefer to take the following approach:
1. Merge these changes so this PR does not remain pending
2. Implement or have another engineer implement these changes in JS and KT
3. Regenerate and validate the test vectors in question
4. Open a new PR against tbdex-js to add the vectors.

# Commits
| issue                                                  | tbdex commits                                                                                                                                                                                                              | tbdex-js commits |
| :----------------------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :----------------------------------------------------- |
| [260](https://github.com/TBD54566975/tbdex/issues/260) | [f55c588](https://github.com/TBD54566975/tbdex/pull/269/commits/f55c588c36f427c42172312a8d99bf1c3b340d37), [752aa4f](https://github.com/TBD54566975/tbdex/pull/269/commits/752aa4fc29fd80e0d8609fe41f97388f7c1e29eb) | [87c4162](https://github.com/TBD54566975/tbdex-js/commit/87c4162288fd6a378925ac181b4c6aed4df41c1f) |
| [261](https://github.com/TBD54566975/tbdex/issues/261) | [9dc5148](https://github.com/TBD54566975/tbdex/commit/9dc514816043ddf8fa7c6cc57b512fc036de48ea)                                                                                                                      | [1d82302](https://github.com/TBD54566975/tbdex-js/commit/1d8230201fcee9db319c2602b731d163d93030c3) |
| [262](https://github.com/TBD54566975/tbdex/issues/262) | [ff03448](https://github.com/TBD54566975/tbdex/commit/ff03448e61ea3db6dff96472551f030d1930a677)                                                                                                                      | [85636b3](https://github.com/TBD54566975/tbdex-js/commit/85636b38de757259cf6ba012675e2a0610064aa3) |
| [263](https://github.com/TBD54566975/tbdex/issues/263) | [0614cf5](https://github.com/TBD54566975/tbdex/commit/0614cf575b465581137908312c6e37c31864391c)                                                                                                                      | [80da7b](https://github.com/TBD54566975/tbdex-js/commit/80da7b668029f6ce7cec540c7d8e07c178234464) |
| [253](https://github.com/TBD54566975/tbdex/issues/253) | [5c1e67a](https://github.com/TBD54566975/tbdex/commit/5c1e67a194938d4f5c40f172e7384d370966b4a4)                                                                                                                      | [fceb7aca](https://github.com/TBD54566975/tbdex-js/commit/fceb7aca0876da0b3f22482aaab3af6617c49971) |

### DIFF
--- a/packages/protocol/build/compile-validators.js
+++ b/packages/protocol/build/compile-validators.js
@@ -49,4 +49,4 @@ const moduleCode = standaloneCode(validator)
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
 
 await mkdirp(path.join(__dirname, '../generated'))
-fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.js'), moduleCode)
+fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.cjs'), moduleCode)
--- a/packages/protocol/build/compile-validators.js
+++ b/packages/protocol/build/compile-validators.js
@@ -45,8 +45,26 @@ for (const schemaName in schemas) {
   validator.addSchema(schemas[schemaName], schemaName)
 }
 
-const moduleCode = standaloneCode(validator)
+const generatedCode = standaloneCode(validator)
+
+// https://github.com/ajv-validator/ajv/issues/2209
+// ESM generation is broken in AJV standalone.
+// In particular, it will "require" files from AJVs runtime directory instead of "import"ing.
+function replaceRequireWithImport(inputString) {
+  const variableNameRegex = /\w+/; // Matches the variable name
+  const moduleNameRegex = /[^"']+/; // Matches the module name
+  const regex = new RegExp(
+      `const\\s+(${variableNameRegex.source})\\s*=\\s*require\\s*\\(\\s*[\"'](${moduleNameRegex.source})[\"']\\s*\\)\\.default`,
+      'g'
+  );
+
+  const replacedString = inputString.replace(regex, 'import { default as $1 } from "$2.js"');
+  return replacedString;
+}
+const moduleCode = replaceRequireWithImport(generatedCode)
+
+
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
 
 await mkdirp(path.join(__dirname, '../generated'))
-fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.cjs'), moduleCode)
+fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.js'), moduleCode)
--- a/packages/protocol/src/dev-tools.ts
+++ b/packages/protocol/src/dev-tools.ts
@@ -43,65 +43,65 @@ export class DevTools {
    */
   static createOfferingData(): OfferingData {
     return {
-      description   : 'Selling BTC for USD',
-      payinCurrency : {
+      description : 'Selling BTC for USD',
+      payin       : {
         currencyCode : 'USD',
-        minAmount    : '0.0',
-        maxAmount    : '999999.99',
+        min          : '0.0',
+        max          : '999999.99',
+        methods      : [{
+          kind                   : 'DEBIT_CARD',
+          requiredPaymentDetails : {
+            $schema    : 'http://json-schema.org/draft-07/schema',
+            type       : 'object',
+            properties : {
+              cardNumber: {
+                type        : 'string',
+                description : 'The 16-digit debit card number',
+                minLength   : 16,
+                maxLength   : 16
+              },
+              expiryDate: {
+                type        : 'string',
+                description : 'The expiry date of the card in MM/YY format',
+                pattern     : '^(0[1-9]|1[0-2])\\/([0-9]{2})$'
+              },
+              cardHolderName: {
+                type        : 'string',
+                description : 'Name of the cardholder as it appears on the card'
+              },
+              cvv: {
+                type        : 'string',
+                description : 'The 3-digit CVV code',
+                minLength   : 3,
+                maxLength   : 3
+              }
+            },
+            required             : ['cardNumber', 'expiryDate', 'cardHolderName', 'cvv'],
+            additionalProperties : false
+          }
+        }]
       },
-      payoutCurrency: {
+      payout: {
         currencyCode : 'BTC',
-        maxAmount    : '999526.11'
+        max          : '999526.11',
+        methods      : [{
+          kind                   : 'BTC_ADDRESS',
+          requiredPaymentDetails : {
+            $schema    : 'http://json-schema.org/draft-07/schema',
+            type       : 'object',
+            properties : {
+              btcAddress: {
+                type        : 'string',
+                description : 'your Bitcoin wallet address'
+              }
+            },
+            required             : ['btcAddress'],
+            additionalProperties : false
+          }
+        }]
       },
       payoutUnitsPerPayinUnit : '0.00003826',
-      payinMethods            : [{
-        kind                   : 'DEBIT_CARD',
-        requiredPaymentDetails : {
-          $schema    : 'http://json-schema.org/draft-07/schema',
-          type       : 'object',
-          properties : {
-            cardNumber: {
-              type        : 'string',
-              description : 'The 16-digit debit card number',
-              minLength   : 16,
-              maxLength   : 16
-            },
-            expiryDate: {
-              type        : 'string',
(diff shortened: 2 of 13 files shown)


packages/protocol/build/compile-validators.js
```diff
@@ -49,4 +49,4 @@ const moduleCode = standaloneCode(validator)
 const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
 
 await mkdirp(path.join(__dirname, '../generated'))
-fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.js'), moduleCode)
\ No newline at end of file
+fs.writeFileSync(path.join(__dirname, '../generated/compiled-validators.cjs'), moduleCode)
```
diehuxx, line 52: When I updated the `tbdex` submodule to bring in the changes, I got a warning when running tbdex-js tests that `compiled-validators.js` was being read as ES and that was causing error. It fixed when I switched this to `cjs`. Seeking feedback if this is not the right thing to do here.

KendallWeihe (approved):
💯

packages/protocol/src/dev-tools.ts
```diff
@@ -43,65 +43,66 @@ export class DevTools {
...
-        minAmount    : '0.0',
-        maxAmount    : '999999.99',
+        min          : '0.0',
+        max          : '999999.99',
+        methods      : [{
+          kind                   : 'DEBIT_CARD',
+          requiredPaymentDetails : {
+            $schema    : 'http://json-schema.org/draft-07/schema',
+            type       : 'object',
+            properties : {
```
nitro-neal, line 56: do you normally need billing address for this?

packages/protocol/build/compile-validators.js
nitro-neal: @frankhinek would know for sure, but I think if the is the only place it could be a problem and not work in certain environments ?

packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -106,28 +106,28 @@ export class Rfq extends Message {
 
     // Verifyin payin amount is less than maximum
```
nitro-neal, line 107:
```suggestion
    // Verifying payin amount is less than maximum
```

packages/protocol/src/validator.ts
```diff
@@ -1,6 +1,6 @@
 import type { ErrorObject } from 'ajv'
 // validator functions are compiled at build time. check ./build/compile-validators.js for more details
-import * as compiledValidators from '../generated/compiled-validators.js'
+import * as compiledValidators from '../generated/compiled-validators.cjs'
```
nitro-neal, line 3: only place in the code where we import a cjs right?

Maybe as a workaround / fix we can just make a normal .ts validator file? Why do we have to build a dynamic validation file?

nitro-neal (approved):
Looks good! Just double check if that cjs is all good to go

packages/protocol/src/dev-tools.ts
```diff
@@ -43,65 +43,66 @@ export class DevTools {
...
-            }
+            required             : ['btcAddress'],
+            additionalProperties : false
           },
-          required             : ['btcAddress'],
-          additionalProperties : false
-        }
-      }],
-      requiredClaims: {
+          estimatedSettlementTime: 10,
```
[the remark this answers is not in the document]
diehuxx: It's in seconds. I can add a comment

packages/protocol/tests/test-vectors.spec.ts
```diff
@@ -22,7 +22,7 @@ describe('TbdexTestVectorsProtocol', function () {
     expect(close.toJSON()).to.deep.eq(ParseClose.output)
   })
 
-  it('parse_offering', async() => {
+  it.skip('parse_offering', async() => {
```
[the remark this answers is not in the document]
diehuxx: There are no up-to-date test vectors for this test. We can't create functioning test vectors until 2 implementations have implemented these changes. I would prefer to take the following approach:
1. merge these changes so this PR does not remain pending
2. implement or have another engineer implement these changes in JS and KT
3. Regenerate and validate the test vectors in question
4. Open a new PR against tbdex-js to add the vectors.

packages/protocol/tests/test-vectors.spec.ts
```diff
@@ -70,7 +70,7 @@ describe('TbdexTestVectorsProtocol', function () {
     expect(quote.toJSON()).to.deep.eq(ParseQuote.output)
   })
 
-  it('parse_rfq', async () => {
+  it.skip('parse_rfq', async () => {
```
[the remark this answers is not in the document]
diehuxx: Ditto https://github.com/TBD54566975/tbdex-js/pull/213#discussion_r1539730516

diehuxx:
> You either need to transform mixed syntax or use cjs or other. There are avenues

Which approach/avenue do you recommend?

packages/protocol/src/dev-tools.ts
diehuxx: I think in a real-life card payment the PFI might need the billing address, but for the purposes of this dummy test data it doesn't matter either way.

packages/protocol/src/validator.ts
mistermoe: @nitro-neal not all runtimes allow you to read files. this is especially the case in constrained CSP environments

diehuxx: Yep it's the only cjs import that I know of.

> Maybe as a workaround / fix we can just make a normal .ts validator file?

I don't think we can, though there may be a neckbeard magic way. We're generating `compiled-validators.*s` using `ajv`'s `standaloneCode` function. We're setting the `esm: true` option when generating the validation code, but it's still spitting out CJS for whatever reason. If we change the file extension to `.ts`, then that would technically be valid `.ts` since typescript is a superset of javascript. But the typescript compiler only knows how to transpile `compiled-validators.ts` into CJS. It can't turn that CJS-ts into ESM AFAIU.

> Why do we have to build a dynamic validation file?


There's a few benefits to doing it this way, laid out in the `ajv` documentation [here](https://ajv.js.org/standalone.html#validating-a-single-schemas-using-the-js-library-es6-and-cjs). But the gist is it's better for browsers to do this work at build time.

diehuxx:
> > Which approach/avenue do you recommend? Does the approach I took fall into the category of using cjs or are you referring another possible approach?
> 
> Yes this approach falls in the category of using cjs.

Which approach/avenue you recommend?

packages/protocol/tests/test-vectors.spec.ts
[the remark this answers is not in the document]
diehuxx: Fair enough. Added.

packages/protocol/src/dev-tools.ts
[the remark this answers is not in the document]
diehuxx: Added comment

mistermoe:
@diehuxx  we've configured ajv to output esm which y'all can see [here](https://github.com/TBD54566975/tbdex-js/blob/main/packages/protocol/build/compile-validators.js#L42). if ajv is returning some cjs that means there's a bug in their codegen. have we checked ajv github issues?

mistermoe:
@diehuxx i would build `dist` and then set up a test script that consumes what's built to make sure that there's no issues. if all good on that front, i'd merge this and create an issue to tackle the ajv thing later.

diehuxx:
> @diehuxx we've configured ajv to output esm which y'all can see [here](https://github.com/TBD54566975/tbdex-js/blob/main/packages/protocol/build/compile-validators.js#L42). if ajv is returning some cjs that means there's a bug in their codegen. have we checked ajv github issues?

Good call. There's an open issue in `ajv` about esm generation being broken. https://github.com/ajv-validator/ajv/issues/2209, but looks like a couple people posted workarounds in the comments. I'm going to give those a try.

diehuxx:
Just pushed a brute force solution to the ESM compiled validators issue. I added some lines to `packages/protocol/build/compile-validators.js` which convert all `require`s into `import`s.

PR #215, opened 2024-03-28
Create salted hashes of privateData in Rfq

diehuxx:
Addresses spec change in [this PR](https://github.com/TBD54566975/tbdex/pull/294)

In future PRs it may be useful to add finer grained controls over keeping/omitting certain privateData fields from serialization or validation, but this PR lays out everything needed to get this breaking change over the line.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,4 +1,4 @@
-name: Security Scanning
+name: Security and License Scans
 
 on:
   pull_request_target:
@@ -17,22 +17,6 @@ on:
   workflow_dispatch:
 
 jobs:
-  fossa:
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          ref: ${{ github.head_ref }}
-
-      - name: Run FOSSA Scan
-        uses: fossas/fossa-action@main
-        with:
-          api-key: ${{ secrets.FOSSA_API_KEY }}
-
-      - name: Run FOSSA Test
-        uses: fossas/fossa-action@main
-        with:
-          api-key: ${{ secrets.FOSSA_API_KEY }}
-          run-tests: true
+  security-license-scan:
+    uses: TBD54566975/open-source-programs/.github/workflows/security.yml@main
+    secrets: inherit
--- a/packages/protocol/src/message-kinds/rfq.ts
+++ b/packages/protocol/src/message-kinds/rfq.ts
@@ -140,17 +140,19 @@ export class Rfq extends Message {
     const { paymentDetails: payinDetails, ...remainingPayin } = payin
     const { paymentDetails: payoutDetails, ...remainingPayout } = payout
 
-    const data = {
+    const data: RfqData = {
       ...remainingRfqData,
       payin        : remainingPayin,
       payout       : remainingPayout,
-      claimsHashes : claims.map((claim) => Rfq.digestPrivateData(salt, claim))
     }
     if (payinDetails !== undefined) {
-      data.payin.paymentDetailsHash = Rfq.digestPrivateData(salt, payinDetails)
+      data.payin!.paymentDetailsHash = Rfq.digestPrivateData(salt, payinDetails)
     }
     if (payoutDetails !== undefined) {
-      data.payout.paymentDetailsHash = Rfq.digestPrivateData(salt, payoutDetails)
+      data.payout!.paymentDetailsHash = Rfq.digestPrivateData(salt, payoutDetails)
+    }
+    if (claims !== undefined && claims?.length > 0) {
+      data.claimsHash = Rfq.digestPrivateData(salt, claims)
     }
 
     const privateData: RfqPrivateData = {
@@ -191,8 +193,8 @@ export class Rfq extends Message {
     }
 
     // Verify claims
-    if (this.data.claimsHashes.length > 0) {
-      this.verifyClaimsHashes()
+    if (this.data.claimsHash !== undefined) {
+      this.verifyClaimsHash()
     }
   }
 
@@ -213,8 +215,8 @@ export class Rfq extends Message {
     }
 
     // Verify claims
-    if (this.data.claimsHashes.length > 0  && this.privateData?.claims !== undefined) {
-      this.verifyClaimsHashes()
+    if (this.data.claimsHash !== undefined  && this.privateData?.claims !== undefined) {
+      this.verifyClaimsHash()
     }
   }
 
@@ -248,22 +250,20 @@ export class Rfq extends Message {
     }
   }
 
-  private verifyClaimsHashes(): void {
+  private verifyClaimsHash(): void {
     if (this?.privateData?.salt === undefined) {
-      throw new Error('Salt must be present to verify data.claimsHashes')
+      throw new Error('Salt must be present to verify data.claimsHash')
     }
 
-    for (let i = 0; i < this.data.claimsHashes.length; i++) {
-      const claimsHash = this.data.claimsHashes[i]
-      const claim = this.privateData?.claims?.[i]
-      const digest = Rfq.digestPrivateData(this.privateData.salt, claim)
+    const claimsHash = this.data.claimsHash!
+    const claims = this.privateData?.claims
+    const digest = Rfq.digestPrivateData(this.privateData.salt, claims)
 
-      if (digest !== claimsHash) {
-        throw new Error(
-          'Private data integrity check failed: ' +
-          `data.claimsHashes[${i}] does not match digest of privateData.claims[${i}]`
-        )
-      }
+    if (digest !== claimsHash) {
+      throw new Error(
+        'Private data integrity check failed: ' +
+        `data.claimsHash does not match digest of privateData.claims`
+      )
     }
   }
 
--- a/packages/protocol/src/message-kinds/rfq.ts
+++ b/packages/protocol/src/message-kinds/rfq.ts
@@ -221,11 +221,7 @@ export class Rfq extends Message {
   }
 
   private verifyPayinDetailsHash(): void {
-    if (this?.privateData?.salt === undefined) {
-      throw new Error('Salt must be present to verify data.payin.paymentDetailsHash')
-    }
-
-    const digest = Rfq.digestPrivateData(this.privateData.salt, this.privateData?.payin?.paymentDetails)
+    const digest = Rfq.digestPrivateData(this.privateData!.salt, this.privateData?.payin?.paymentDetails)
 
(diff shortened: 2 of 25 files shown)


packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -72,18 +112,172 @@ export class Rfq extends Message {
...
 
+  /**
+   * Hash private RFQ data and set private fields in an RfqPrivateData object
+   * @param unhashedRfqData
+   * @returns An object with fields data and privateData.
+   * @returns {@link RfqData} The value of data field.
+   * @returns {@link RfqPrivateData} The value of privateData field.
+   */
+  private static hashPrivateData(unhashedRfqData: UnhashedRfqData): { data: RfqData, privateData: RfqPrivateData } {
+    const salt = '123' // TODO: generate randomly
```
diehuxx, line 131: oh lol woops. lemme fix that
- [ ] randomly generate salt

packages/protocol/src/message-kinds/rfq.ts
```diff
@@ -171,7 +380,7 @@ export class Rfq extends Message {
       } else {
         // requiredPaymentDetails is present, so Rfq's payment details must match
         const validate = ajv.compile(paymentMethodMatch.requiredPaymentDetails)
-        const isValid = validate(rfqPaymentMethod.paymentDetails)
+        const isValid = validate(selectedPaymentDetails)
         if (isValid) {
           // Selected payment method matches one of the offering's allowed payment methods
           return
```
jiyoontbd, line 386: i think i might be hallucinating but i want to double check - i believe in line 392 

```js
throw new Error(
      `rfq ${payDirection}Method paymentDetails could not be validated against offering requiredPaymentDetails. ` +
      `Schema validation errors: ${Array.from(invalidPaymentDetailsErrors).join()}`
    )
```
where an error is thrown, that's done without checking if invalidPaymentDetailsErrors is empty. 

meaning, this method _always_ throws an error regardless of whether the validator check passes or fails?

should the error throwing be wrapped in an `if error set is empty` check?

please feel free to tell me i'm wrong. it's late and my eyes are probably failing me

packages/protocol/tests/rfq.spec.ts
```diff
@@ -114,6 +114,116 @@ describe('Rfq', () => {
 
       expect(jsonMessage).to.equal(JSON.stringify(parsedMessage))
     })
+
+    describe('requireAllPrivateData: true', () => {
```
jiyoontbd, line 118: would you add a `describe('requireAllPrivateData: false'..` so we can test `verifyPresentPrivateData` in the else branch?

```js
   if (opts?.requireAllPrivateData) {
      rfq.verifyAllPrivateData()
    } else {
      rfq.verifyPresentPrivateData()
    }
```

i know it uses the same underlying methods, but just a couple test cases to prove we're hitting the else branch would be good :)

tbdex
jiyoontbd, line 1: do you have a PR for these changes in the submodule?

packages/protocol/src/types.ts
```diff
@@ -238,15 +240,43 @@ export type MessageData = RfqData | QuoteData | OrderData | OrderStatusData | Cl
...
+  claimsHashes: string[]
+}
+
+/**
+ * Private data contained in a RFQ message
+ * @beta
+ */
+export type RfqPrivateData = {
+  /** Randomly generated cryptographic salt used to hash privateData fields */
+  salt?: string
```
KendallWeihe, line 257: Should this maybe be a required property?

packages/protocol/src/types.ts
```diff
@@ -238,15 +240,43 @@ export type MessageData = RfqData | QuoteData | OrderData | OrderStatusData | Cl
...
   /** claims that fulfill the requirements declared in an Offering */
+  claims?: string[]
+  /** Selected payment method that Alice will use to send the listed payin currency to the PFI. */
+  payin?: PrivatePaymentDetails
+  /** Selected payment method that the PFI will use to send the listed base currency to Alice */
+  payout?: PrivatePaymentDetails
+}
+
+export type UnhashedRfqData = Omit<RfqData, 'payin' | 'payout' | 'claimsHashes'> & {
```
KendallWeihe, line 266: do we need doc comments on & in this type?

packages/protocol/src/types.ts
```diff
@@ -238,15 +240,43 @@ export type MessageData = RfqData | QuoteData | OrderData | OrderStatusData | Cl
...
+  payin: Omit<SelectedPayinMethod, keyof PrivatePaymentDetails> & PrivatePaymentDetails
+  payout: Omit<SelectedPayoutMethod, keyof PrivatePaymentDetails> & PrivatePaymentDetails,
   claims: string[]
 }
 
+export type PrivatePaymentDetails = {
```
KendallWeihe, line 272: doc comment?

packages/protocol/src/types.ts
KendallWeihe: oh I see you're doing something like this in the verify functions

```ts
    if (this?.privateData?.salt === undefined) {
      throw new Error('Salt must be present to verify data.payout.paymentDetailsHash')
    }
```

hmmm the joys of typescript at buildtime vs javascript at runtime 🤔

packages/protocol/tests/generate-test-vectors.ts
```diff
@@ -104,6 +104,55 @@ const generateParseRfqVector = async () => {
   }
 }
 
+const generateParseRfqOmitPrivateDataVector = async () => {
+  const vc = await VerifiableCredential.create({
+    type    : 'PuupuuCredential',
```
KendallWeihe, line 109: is it 💩 💩? or is it 🔫 🔫 ? what's the difference?

```suggestion
    type    : 'PewPewCredential',
```

packages/protocol/tests/generate-test-vectors.ts
```diff
@@ -104,6 +104,55 @@ const generateParseRfqVector = async () => {
...
+    metadata : { from: aliceDid.uri, to: pfiDid.uri,  protocol: '1.0' },
+    data     : {
+      offeringId : Resource.generateId('offering'),
+      payin      : {
+        kind           : 'DEBIT_CARD',
+        amount         : '20000.00',
+        paymentDetails : {
+          'cardNumber'     : '1234567890123456',
+          'expiryDate'     : '12/22',
+          'cardHolderName' : 'Ephraim Bartholomew Winthrop',
```
KendallWeihe, line 129: hopefully Ephraim is cool with us using their card info for our tests. thanks Ephraim!

KendallWeihe:
some comments but this is looking great! good job @jiyoontbd for the tests things too, b/c I would've missed those

packages/protocol/src/message-kinds/rfq.ts
diehuxx: The function `return`s early in the `for` loop if the payment details are valid, so we should only get to the `throw new Error(` if none of the Offering's payment methods work with the Rfq's `paymentDetails`.

It "works" as-is, but lemme play around with it to make it more readable, bcuz i agree it's a confusingly written function.

packages/protocol/src/types.ts
diehuxx: > hmmm the joys of typescript at buildtime vs javascript at runtime 🤔

lol, you get it. I was on the fence about whether to make this a required typescript property. Since it's "required" for practical purposes, i decided to leave it as an optional field that we check at runtime. I don't feel strongly either way, so just say the word and I'll make it required

tbdex
diehuxx: Yep, this PR in the `tbdex` repo: https://github.com/TBD54566975/tbdex/pull/294.
once that lands, I'll update the submodule again to `tbdex` `origin/main`.

packages/protocol/tests/rfq.spec.ts
```diff
@@ -114,6 +114,116 @@ describe('Rfq', () => {
...
+      it('throws if salt is missing but hashed fields are present in Rfq.data', async () => {
+        const aliceDid = await DidJwk.create()
+        const rfq = Rfq.create({
+          metadata : { from: aliceDid.uri, to: 'did:ex:pfi' },
+          data     : await DevTools.createRfqData()
+        })
+        await rfq.sign(aliceDid)
+
+        const rfqJson = rfq.toJSON()
+        delete rfqJson.privateData!.salt
```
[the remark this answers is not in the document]
diehuxx: > it's a bit scary that js can do this.

tru. 

> but that also means we can't test for this case in kt? 🤔

yea i haven't decided how to write this test in KT yet. I'll let you know lol.

packages/protocol/tests/rfq.spec.ts
```diff
@@ -114,6 +114,116 @@ describe('Rfq', () => {
...
+
+        try {
+          await Rfq.parse(rfqJson, { requireAllPrivateData: true })
+          expect.fail()
+        } catch(e) {
+          expect(e.message).to.include('data.payout.paymentDetailsHash does not match digest of privateData.payout.paymentDetails')
+        }
+      })
+
+      it('throws if Rfq.privateData.claims is missing but Rfq.data.claimsHashes is present', async () => {
```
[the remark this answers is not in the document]
diehuxx: Good call. Added

packages/protocol/tests/rfq.spec.ts
diehuxx: added

packages/protocol/src/types.ts
diehuxx: done

packages/protocol/src/types.ts
diehuxx: addedd

.github/workflows/security.yml
```diff
@@ -1,4 +1,4 @@
-name: Security Scanning
+name: Security and License Scans
```
jiyoontbd, line 1: i think this is leo's changes? do you need to merge from main?

jiyoontbd (approved):

.github/workflows/security.yml
diehuxx: ope my b. murged.

packages/protocol/src/types.ts
KendallWeihe: I think the proper thing is to make the type reflect the data requirements because as a developer when I'm using that type I expect it to reflect the type in the specification, but that may result in some weird ts compiler thing where you still want to check if it's `undefined` even tho from ts's perspective it never can be

KendallWeihe (approved):
I think make the `salt` property of type `string` instead of `string | undefined` as long as it doesn't introduce bonkers typescript side effects

great work! this was no easy lift 😌

PR #217, opened 2024-03-28
Rip out List Offering query params

diehuxx:
Addresses spec change: https://github.com/TBD54566975/tbdex/pull/302

### DIFF
--- a/packages/http-server/src/request-handlers/get-offerings.ts
+++ b/packages/http-server/src/request-handlers/get-offerings.ts
@@ -1,5 +1,5 @@
 import { Request, Response } from 'express'
-import type { GetOfferingsCallback, GetOfferingsFilter, OfferingsApi } from '../types.js'
+import type { GetOfferingsCallback, OfferingsApi } from '../types.js'
 
 type GetOfferingsOpts = {
   callback?: GetOfferingsCallback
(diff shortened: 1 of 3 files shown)


jiyoontbd (approved):
goodbye filters 👋

packages/http-server/src/request-handlers/get-offerings.ts
```diff
@@ -9,20 +9,12 @@ type GetOfferingsOpts = {
...
-  }
-
-  const offerings = await offeringsApi.getOfferings({ filter })
+  const offerings = await offeringsApi.getOfferings()
 
   if (callback) {
     // TODO: figure out what to do with callback result. should we pass through the offerings we've fetched
     //       and allow the callback to modify what's returned? (issue #11)
-    await callback({ request, response }, filter)
+    await callback({ request, response }, {})
```
KendallWeihe, line 17: checkout the `types.ts` file there are `GetOfferingsCallback` and `GetOfferingsFilter` which we can remove too, tho technically this works

KendallWeihe: oh I see, maybe leave it in because we expect to add some in due time

KendallWeihe (approved):

packages/http-server/src/request-handlers/get-offerings.ts
KendallWeihe: you could at least blank this type out to be empty

```ts
export type GetOfferingsFilter = {
  /** Currency that the PFI is buying in exchange for payout currency - ISO 3166 currency code string */
  payinCurrency?: string

  /** Currency that the PFI is selling - ISO 3166 currency code string */
  payoutCurrency?: string

  /** The payin method used to pay money to the PFI */
  payinMethodKind?: string

  /** The payout method to receive money from the PFI  */
  payoutMethodKind?: string

  /** Offering ID */
  id?: string
}
```

diehuxx: good call. Done

PR #219, opened 2024-03-28
Add Close.success

diehuxx:


wesbillman (approved):

packages/protocol/src/types.ts
```diff
@@ -342,5 +342,6 @@ export type OrderStatusData = {
 export type CloseData = {
   /** an explanation of why the exchange is being closed */
   reason?: string
+  /** indicates whether or not the exchange successfully completed */
+  success?: string
```
angiejones, line 346: should this be boolean?

jiyoontbd (approved):

KendallWeihe:
@diehuxx I think we also need to update the `message-kinds/order.ts` in the `protocol` package to allow `close` messages after an `order` message, right?

```ts
readonly validNext = new Set<MessageKind>(['orderstatus'])
```

PR #234, opened 2024-04-05
Boosted http-client code coverage to 100% + minor updates

thehenrytsai:
- Boosted `http-client` to have 100% code coverage
- Removed seemingly unnecessary explicit calls of `setPrototypeOf()`
- Added `html` reporter for code coverage for local detailed report

### DIFF
--- a/packages/http-client/tests/client.spec.ts
+++ b/packages/http-client/tests/client.spec.ts
@@ -486,25 +486,24 @@ describe('client', () => {
 
     it('returns array of message exchanges given a list of exchange IDs', async () => {
       const alice = await DidJwk.create()
-      const aliceRfq = await DevTools.createRfq({ sender: alice })
-      await aliceRfq.sign(alice)
+      const aliceRfq1 = await DevTools.createRfq({ sender: alice })
+      await aliceRfq1.sign(alice)
 
-      const bob = await DidJwk.create()
-      const bobRfq = await DevTools.createRfq({ sender: bob })
-      await bobRfq.sign(bob)
+      const aliceRfq2 = await DevTools.createRfq({ sender: alice })
+      await aliceRfq2.sign(alice)
 
       fetchStub.resolves({
         ok   : true,
         json : () => Promise.resolve({ data: [
-          [aliceRfq.toJSON()],
-          [bobRfq.toJSON()]
+          [aliceRfq1.toJSON()],
+          [aliceRfq2.toJSON()]
         ] })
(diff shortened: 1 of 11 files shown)


packages/http-client/tests/client.spec.ts
```diff
@@ -475,14 +483,30 @@ describe('client', () => {
...
-    it('returns empty exchanges array if response is ok and body is empty array', async () => {
+
+    it('returns array of message exchanges given a list of exchange IDs', async () => {
+      const alice = await DidJwk.create()
+      const aliceRfq = await DevTools.createRfq({ sender: alice })
+      await aliceRfq.sign(alice)
+
+      const bob = await DidJwk.create()
+      const bobRfq = await DevTools.createRfq({ sender: bob })
+      await bobRfq.sign(bob)
```
jiyoontbd, line 494: maybe what would make more sense naming wise is to say `aliceRfq1` and `aliceRfq2`, since the GET /exchanges call would be returning all exchanges for a given did? so maybe something like 

```suggestion
      const aliceRfq2 = await DevTools.createRfq({ sender: alice })
      await aliceRfq2.sign(alice)
```

packages/http-client/tests/client.spec.ts
```diff
@@ -475,14 +483,30 @@ describe('client', () => {
       }
     })
 
-    it('returns empty exchanges array if response is ok and body is empty array', async () => {
+
+    it('returns array of message exchanges given a list of exchange IDs', async () => {
+      const alice = await DidJwk.create()
+      const aliceRfq = await DevTools.createRfq({ sender: alice })
+      await aliceRfq.sign(alice)
```
jiyoontbd, line 490:
```suggestion
      const aliceRfq1 = await DevTools.createRfq({ sender: alice })
      await aliceRfq1.sign(alice)
```

jiyoontbd (approved):

packages/http-client/tests/client.spec.ts
thehenrytsai: Done. Thanks @jiyoontbd for the knowledge, makes sense, didn't realize the `did` property was meant to be for the caller, I've updated that too, let me know if I misunderstood again!

packages/http-client/tests/client.spec.ts
thehenrytsai: Done.

diehuxx (approved):

Issue #239, opened 2024-04-09
Return array of `id`s from getExchanges per spec change


kirahsapong:
Per spec change: https://github.com/TBD54566975/tbdex/tree/main/specs/http-api#list-exchanges

cc: @KendallWeihe 

detail here: https://github.com/TBD54566975/tbdex-js/pull/237#discussion_r1558061279

And remove query params altogether.

PR #240, opened 2024-04-09
Fixed bug where IDs passed to getExchanges is handled incorrectly

thehenrytsai:
1. Fixed bug where IDs passed to getExchanges is handled incorrectly thus you cannot get exchanges given a list of IDs
2. Fixed bug in the test that is meant to test this (it always passes due to incorrect usage)
3. Fixed same issue in another test, switched to `sinon-chai` to give better readability and debug error message
4. 100% code coverage on `create-exchange.ts`

### DIFF
--- a/.changeset/fair-spiders-eat.md
+++ b/.changeset/fair-spiders-eat.md
@@ -2,4 +2,4 @@
 "@tbdex/http-server": patch
 ---
 
-Fixed bug where IDs passed to getExchanges is handled incorrectly.
+Fixed bugs in tests involving calledWith() and increased code coverage.
--- a/.changeset/fair-spiders-eat.md
+++ b/.changeset/fair-spiders-eat.md
@@ -1,0 +1,5 @@
+---
+"@tbdex/http-server": patch
+---
+
+Fixed bug where IDs passed to getExchanges is handled incorrectly.
--- a/packages/http-server/tests/create-exchange.spec.ts
+++ b/packages/http-server/tests/create-exchange.spec.ts
@@ -1,7 +1,7 @@
 import { ErrorDetail, Offering, Order, Rfq } from '@tbdex/http-client'
 import type { Server } from 'http'
 
-import { DevTools, RequestContext, TbdexHttpServer } from '../src/main.js'
+import { CallbackError, DevTools, RequestContext, TbdexHttpServer } from '../src/main.js'
 import { expect } from 'chai'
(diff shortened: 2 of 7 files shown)


mistermoe:
Nice @thehenrytsai. Wondering if we should go ahead and change `GET /exhanges` to snap to what's defined in the spec in a subsequent PR soon.

thehenrytsai:
> Nice @thehenrytsai. Wondering if we should go ahead and change `GET /exhanges` to snap to what's defined in the spec in a subsequent PR soon.

Hey @mistermoe, are you referring to #239? Assuming so, sanity:

So there is no (longer a) need to even pass IDs as query param? If so, the implementation of supporting multiple IDs is seems completely redundant. Am I misunderstanding something?

.changeset/fair-spiders-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tbdex/http-server": patch
+---
+
+Fixed bug where IDs passed to getExchanges is handled incorrectly.
```
KendallWeihe, line 5: Looks like we undid this during the life of this PR, yeah? If so, can we update the language to reflect changes

packages/http-server/tests/create-exchange.spec.ts
```diff
@@ -342,6 +342,44 @@ describe('POST /exchanges/:exchangeId/rfq', () => {
       expect(lastCallbackArg.replyTo).to.be.undefined
     })
 
+    it('propagates the status code and detail if custom handler throws a custom error', async () => {
```
KendallWeihe, line 345: nice, these tests are 💯

KendallWeihe (approved):
please update the changeset language to reflect the latest, but other than that ✅

.changeset/fair-spiders-eat.md
thehenrytsai: Good call! Updated.

mistermoe:
> > Nice @thehenrytsai. Wondering if we should go ahead and change `GET /exhanges` to snap to what's defined in the spec in a subsequent PR soon.
> 
> 
> 
> Hey @mistermoe, are you referring to #239? Assuming so, sanity:
> 
> 
> 
> So there is no (longer a) need to even pass IDs as query param? If so, the implementation of supporting multiple IDs is seems completely redundant. Am I misunderstanding something?


That's right @thehenrytsai ! I see no need to pass IDs as a query param and supporting many would be redundant.

PR #245, opened 2024-04-25
fix: use fragment in json schema refs

diehuxx:


mistermoe (approved):

KendallWeihe (approved):

packages/protocol/src/dev-tools.ts
```diff
@@ -87,7 +87,7 @@ export class DevTools {
         methods      : [{
           kind                   : 'BTC_ADDRESS',
           requiredPaymentDetails : {
-            $schema    : 'http://json-schema.org/draft-07/schema',
+            $schema    : 'http://json-schema.org/draft-07/schema#',
```
jiyoontbd, line 90: What does adding # here do?

jiyoontbd (approved):

packages/protocol/src/dev-tools.ts
thehenrytsai: Seems to be for better compatibility between JSON schema processor implementations.

Issue #252, opened 2024-06-12
Investigate why `TbdexTestVectorsProtocol` tests runs extremely slow sometimes and fix it


thehenrytsai:
`TbdexTestVectorsProtocol` tests that involves JSON parsing SOMETIMES take insane amount of time (~5 seconds per test) to complete causing test timeout failures.

Glancing over the code however I don't see why it'd take milliseconds to complete sometimes and 10x or even 100x some other time. But this causes our CICD pipeline to fail with a misleading error because the test log are hidden when run using `pnpm` (there might be a config to expand it, but that's a separate issue):

<img width="1055" alt="image" src="https://github.com/TBD54566975/tbdex-js/assets/17891086/680fd6a5-b106-48a6-8023-48b9c2488b7f">

michaelneale:
and it is fine when running locally?

thehenrytsai:
@michaelneale, this appears must more rarely locally, but I have seen it when I inherited the repo and was very puzzled by it, but never reproduced it twice in a row.

michaelneale:
I think it is the mocha xml reporter that blows up - I saw it locally a bit too, couldn't work out why. XML isn't a great format for streaming results and can blow up in memory but not sure in this case? 

https://github.com/TBD54566975/tbdex-js/pull/255 I have run that several times with no issue with the default spec reporter

michaelneale:
tried updating to latest mocha stuff - same issue.

michaelneale:
certtainly seems related to xml reporting.

thehenrytsai:
I think #257 should do the job.

thehenrytsai:
#257 addresses this, closing.

PR #255, opened 2024-06-14
try different test reporter

michaelneale:
may be a bad mocha test reporter

### DIFF
--- a/.github/workflows/integrity-check.yml
+++ b/.github/workflows/integrity-check.yml
@@ -42,14 +42,8 @@ jobs:
         run: pnpm lint
 
       - name: Run tests for all packages
-        run: pnpm test:node --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
+        run: pnpm test:node
       
-      - name: Upload test results to update SDK reports
-        uses: actions/upload-artifact@v3
-        with:
-          name: junit-results
-          path: packages/*/results.xml
-
       - name: Upload test coverage to Codecov
         uses: codecov/codecov-action@eaaf4bedf32dbdc6b720b63067d99c4d77d6047d # v3.1.4
         env:


michaelneale:
fix for: https://github.com/TBD54566975/tbdex-js/issues/252 and fixes the build

.github/workflows/integrity-check.yml
```diff
@@ -42,14 +42,8 @@ jobs:
         run: pnpm lint
 
       - name: Run tests for all packages
-        run: pnpm test:node --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
+        run: pnpm test:node
       
-      - name: Upload test results to update SDK reports
```
jiyoontbd, line 47: what are the consequences if we get rid of this step? i feel like either @leordev or @finn-tbd set this up for some dashboard view

jiyoontbd: i want to say it's related to this page https://tbd54566975.github.io/sdk-report-runner/ but could be totally wrong

michaelneale: yeah - would like their input, as not sure. Could convert it to XML another way?

michaelneale: I tried updating to latest mocha and reporter, same issue.

jiyoontbd: It could be due to something other than the test runner. The test that flaky fails on CI and local seems to always be the one that parses in the test vector as JSON due to timeout.

michaelneale: yeah it is likely uncovering some other problem.

michaelneale: Ah just saw this locally: 

image

So it isn't really a fix, just hiding some timeout.

michaelneale:
I don't think this fixes it yet BTW< saw this locally: 

image

michaelneale:
closing this as it didn't really solve it. Trying this approach: 

https://github.com/TBD54566975/tbdex-js/pull/256

.github/workflows/integrity-check.yml
finn-tbd: Consequences of getting rid of this step are indeed related to that page. It will no longer be able to display information about the tbdex-js, and just fully removing these steps may fully break the compliance report runner until it's updated to stop checking tbdex-js.

PR #256, opened 2024-06-14
tidy up tests to tell runner it is done

michaelneale:
Trying to clean up after tests a bit as that seemed to help locally

### DIFF
--- a/packages/protocol/tests/test-vectors.spec.ts
+++ b/packages/protocol/tests/test-vectors.spec.ts
@@ -11,100 +11,173 @@ import { Balance, Close, Offering, Order, OrderStatus, Quote, Rfq } from '../src
 import { Parser } from '../src/parser.js'
 
 describe('TbdexTestVectorsProtocol', function () {
-  this.timeout(10000)
-  it('parse_close', async () => {
-    // Parse with parseMessage()
-    const message = await Parser.parseMessage(ParseClose.input)
-    expect(message.isClose()).to.be.true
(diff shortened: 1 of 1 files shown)


packages/protocol/tests/test-vectors.spec.ts
```diff
@@ -11,100 +11,173 @@ import { Balance, Close, Offering, Order, OrderStatus, Quote, Rfq } from '../src
...
-    // Parse with Close.parse()
-    const close = await Close.parse(ParseClose.input)
-    expect(close.isClose()).to.be.true
-    expect(close.toJSON()).to.deep.eq(ParseClose.output)
+  this.timeout(20000) // Increase the timeout to 20 seconds for all tests
+
+  it('parse_close', function (done) {
+    (async () => {
+      try {
+        console.log('Starting parse_close test')
```
jiyoontbd, line 19: I was tinkering with this earlier and chad told me there's something called console.time() and console.timeEnd() you sandwich between a call you want to measure the time of. Might be helpful in your debugging efforts here!

thehenrytsai:
I think the existing timeout of 10 seconds is a bad practice which is probably a bandaid to get things going back in the crunch time of tbDEX v1 shipping mode.

Investigated and provide root cause fix which I believe should be the way forward and these test should NEVER timeout again LOL! Happy to discuss if needed: #257

michaelneale:
closing this as I don't think needed.

PR #257, opened 2024-06-14
#252 - Fixed long running parsing tests due to DID resolution

thehenrytsai:
These tests perform a signature verification which involves resolution of DID DHT DIDs that can take more than 10s from time to time causing the tests to fail.

With this change all these tests now take less than 1 millisecond to run combined.

The intent of these tests seems to be validating the parsing logic, thus I think we don't actually need the real signature verification so we can just mock it.

Test coverage remains the same as other test have hit the signature verification methods.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,7 +9,7 @@
 # The format is described: https://github.blog/2017-07-06-introducing-code-owners/
 
 # These owners will be the default owners for everything in the repo.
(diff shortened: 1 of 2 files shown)


jiyoontbd (approved):

nitro-neal:
Legend

michaelneale:
very nice!

PR #268, opened 2024-08-07
Added OrderInstructions message

thehenrytsai:


ethan-tbd (approved):

jiyoontbd:
@thehenrytsai could you remove payment instructions from quote.data.payin and quote.data.payout?

https://github.com/TBD54566975/tbdex/tree/main/specs/protocol#quote

image

packages/protocol/src/message-kinds/order.ts
```diff
@@ -16,12 +16,14 @@ export type CreateOrderOptions = {
  */
 export class Order extends Message {
   /** a set of valid Message kinds that can come after an order */
-  readonly validNext = new Set<MessageKind>(['orderstatus'])
+  readonly validNext = new Set<MessageKind>(['orderinstructions'])
```
jiyoontbd, line 19: i have these in tbdex-go

```go
// ValidNext returns the valid message kinds that can follow an order.
func ValidNext() []string {
	return []string{orderstatus.Kind, closemsg.Kind, cancel.Kind}
}
```

packages/protocol/tests/exchange.spec.ts
```diff
@@ -258,20 +273,43 @@ describe('Exchange', () => {
...
+      it('can add an OrderInstructions after Order', async () => {
         const exchange = new Exchange()
 
         exchange.addMessages([rfq, quote, order])
 
+        exchange.addNextMessage(orderInstructions)
+        expect(exchange.orderInstructions).to.deep.eq(orderInstructions)
+      })
+
+      it('cannot add Rfq, Quote, Order, OrderStatus, or Close after Order', async () => {
```
jiyoontbd, line 285: technically you should be able to add a close after an order

jiyoontbd (approved):
couple comments to address, but otherwise looks good! thanks henry!

packages/protocol/tests/exchange.spec.ts
thehenrytsai: Thanks, will update!

packages/protocol/src/message-kinds/order.ts
thehenrytsai: Addressed.

packages/protocol/tests/exchange.spec.ts
thehenrytsai: Addressed.

PR #269, opened 2024-08-12
Added Cancel message support

thehenrytsai:


packages/protocol/src/exchange.ts
```diff
@@ -26,6 +26,8 @@ export class Exchange {
   orderstatus: OrderStatus[]
   /** Message sent by either the PFI or Alice to terminate an exchange */
```
angiejones, line 27:
```suggestion
  /** Message sent by the PFI to terminate an exchange */
```

packages/protocol/src/message-kinds/cancel.ts
```diff
@@ -0,0 +1,73 @@
...
+  readonly data: CancelData
+
+  constructor(metadata: CancelMetadata, data: CancelData, signature?: string) {
+    super(metadata, data, signature)
+    this.metadata = metadata
+    this.data = data
+  }
+
+  /**
+   * Parses a JSON message into an Cancel.
```
angiejones, line 38:
```suggestion
   * Parses a JSON message into a Cancel.
```

packages/protocol/src/message-kinds/cancel.ts
```diff
@@ -0,0 +1,73 @@
...
+   * Creates an Cancel with the given options.
+   * @param opts - Options to create an Cancel.
+   */
+  static create(opts: CreateCancelOptions): Cancel {
+    const metadata: CancelMetadata = {
+      ...opts.metadata,
+      kind      : 'cancel',
+      id        : Message.generateId('cancel'),
+      createdAt : new Date().toISOString(),
+      protocol  : opts.metadata.protocol ?? '1.0'
```
angiejones, line 66:
```suggestion
      protocol  : opts.metadata.protocol ?? '2.0'
```

Cancel is introduced in 2.0

packages/protocol/src/message-kinds/close.ts
```diff
@@ -12,17 +12,19 @@ export type CreateCloseOptions = {
 }
 
 /**
- * A Close can be sent by Alice or the PFI as a reply to an RFQ or a Quote
+ * A Close can only be sent the PFI as a reply to an RFQ or a Quote
```
angiejones, line 15:
```suggestion
 * A Close can only be sent the PFI as a reply to an RFQ, Quote, or Cancel
```

packages/protocol/src/message-kinds/close.ts
```diff
@@ -68,9 +70,4 @@ export class Close extends Message {
     close.validateData()
     return close
   }
-
-  /** an explanation of why the exchange is being closed */
-  get reason() {
```
angiejones, line 73: how come this is being removed?

packages/protocol/src/types.ts
```diff
@@ -478,7 +484,7 @@ export enum OrderStatusEnum {
 }
 
 /**
- * A Close can be sent by Alice or the PFI as a reply to an RFQ or a Quote
+ * A Close can only be sent by the PFI at any point during the exchange, to indicate the PFI will not continue the exchange.
```
angiejones, line 487: this implies the exchange is not completed, but it could have completed successfully

```suggestion
 * A Close, which only the PFI can send, can be submitted at any point during the exchange to indicate that the transaction has reached a terminal state.
```

packages/protocol/src/exchange.ts
thehenrytsai: Thanks @angiejones! Done.

packages/protocol/src/message-kinds/cancel.ts
thehenrytsai: Done!

packages/protocol/src/message-kinds/cancel.ts
thehenrytsai: Following spec which uses 1.0 here: https://github.com/TBD54566975/tbdex/tree/main/specs/protocol#example-cancel

packages/protocol/src/message-kinds/close.ts
thehenrytsai: Thanks for the call out, adding tests.

packages/protocol/src/message-kinds/close.ts
thehenrytsai: Doesn't appear to be used, and isn't tested.

packages/protocol/src/types.ts
thehenrytsai: Taken your suggestion. FYI, the language was copied from spec:

https://github.com/TBD54566975/tbdex/tree/main/specs/protocol#close

packages/protocol/src/message-kinds/cancel.ts
angiejones: grrr we really need to get that updated lol

packages/protocol/src/message-kinds/close.ts
angiejones: used by clients to determine why their transaction was closed. it's part of the [spec](https://github.com/TBD54566975/tbdex/tree/main/specs/protocol#close)

angiejones (approved):

packages/protocol/src/message-kinds/close.ts
thehenrytsai: Hey @angiejones thanks for approving! Yeah `reason` can be access already via `close.data.reason`, so extra sugar is not needed (and not tested, even though trivial), hence removal.

nitro-neal (approved):
This is a masterclass PR, thanks a lot great job!

